PR #9, opened 2018-04-16
Xang Thao completed object assignment and stretch

P-web-art:


assignments/arrays.js
```diff
@@ -63,34 +63,57 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+console.log(`Car 33 is a ${inventory[32]["car_year"]} ${inventory[32]["car_make"]} ${inventory[32]["car_model"]}` );
```
enitchals, line 66: Nice use of template literals. They're so powerful!

assignments/arrays.js
```diff
@@ -63,34 +63,57 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
-let lastCar = 0;
-console.log();
+let lastCar = inventory[inventory.length - 1];
+console.log(lastCar["car_make"] + " " + lastCar["car_model"]);
 
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
 let carModels = [];
-console.log();
+carModels = inventory.sort(function(a, b) {
```
enitchals, line 78: This is a sophisticated solution that would be great even for a dataset that was less uniform than the one we have.

assignments/callbacks.js
```diff
@@ -1,34 +1,101 @@
...
+console.log(sumNums(3, 4, sumNumsCallback));
+
+let multiplyNumsCallback = (a, b) => {
+  return a * b;
+};
+
+let multiplyNums = (x, y, cb) => {
   // multiplyNums multiplies two numbers and passes the result to the callback.
-}
+  return cb(x, y);
```
enitchals, line 57: In this line, you're passing both x and y to the callback as two unchanged values rather than multiplying them and passing one result to the callback.

assignments/callbacks.js
```diff
@@ -1,34 +1,101 @@
...
+    if(!nonDuplicates.includes(array[i])) {
+      nonDuplicates.push(array[i]);
+    }
+  }
+  return nonDuplicates;
+};
+
+let test = [1, 2, 5, 2, 6, 1, 3, 7, 1];
+
+let removeDuplicates = (array, cb) => {
```
enitchals, line 93: If you move your code from `removeDuplicatesCallback` into the `removeDuplicates` function, and then pass `nonDuplicates` to the callback in your return statement instead of `array`, this should be good to go :+1:

enitchals (commented):
Great work with this. If you ever want extra stretch problems, please let me know!

PR #10, opened 2018-04-16
Thuy Pham

iamthuypham:


ghost (commented):
Honestly this is barely distinguishable from my own code. In the final example of callbacks however you forgot to use cb and if we're not using callback we might aswell just add items to a set for its default behavior.

Semi colons are an optional component in the wild but I see alot of merit in them and will require them to properly delimit assignments and function calls.

assignments/callbacks.js
```diff
@@ -31,4 +54,13 @@ function removeDuplicates(array, cb) {
   // removeDuplicates removes all duplicate values from the given array.
   // Pass the duplicate free array to the callback function.
   // Do not mutate the original array.
+  let hash = {};
+  let res = []
+  for (let i=0; i<array.length; i++){
+    if (!hash[array[i]]) {
+      res.push(array[i])
+      hash[array[i]] = true
```
ghost, line 62: Isn't this the coolest move?

ghost:
This is definitely a passing example of this project. You did an exellent job and my head spun how quickly you completed it. :+1: :+1: Nothing new to say since yesterday, lets just keep moving to JS-II.

assignments/callbacks.js
iamthuypham: No, it doesn't. `Set` is the best!

ghost: Baring that Set isn't available and we don't want to depend upon its time complexity, this is a solid O(n) capable of providing an early break case should you need one. In this, hashing is the best and you will use this exact code many times in your future.

PR #17, opened 2018-04-16
William VanDolah

wvandolah:


JasonCorp84:
William,

You completed all the assignments and used a different method to declare objects, which is really fancy.
Let's try to do the stretch assignments also.
Keep up the good work!

PR #18, opened 2018-04-16
Rick Espinoza JSI objects

combatpacifist:


JasonCorp84:
Rick,
Good job on the object.js and array.js file. An easier way without using the forEach method would be inventory[id-1]. On the other hand, your code is working perfectly with unknown sequence of ID-s.

combatpacifist:
Thanks!!!!

PR #22, opened 2018-04-16
Jose Garza

jggarza5:
Javascript-I

### DIFF
--- a/assignments/arrays.js
+++ b/assignments/arrays.js
@@ -63,34 +63,54 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+console.log(`Car 33 is a ${inventory[32].car_year} ${inventory[32].car_make} ${inventory[32].car_model}` );
 
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
 let lastCar = 0;
-console.log();
+console.log(inventory[inventory.length - 1].car_make + " " + inventory[inventory.length - 1].car_model);
 
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
 let carModels = [];
-console.log();
+for (let i = 0; i < inventory.length; i++) {
+    carModels.push(inventory[i].car_model);
+}
+carModels.sort();
+console.log(carModels);
+console.log(" ");
 
 // ==== Challenge 4 ====
 // The accounting team needs all the years from every car on the lot. Create a new array from the dealer data containing only the car years and log the result in the console.
 let carYears = [];
-console.log();
+for (let i = 0; i < inventory.length; i++) {
+    carYears.push(inventory[i].car_year);
+}
+console.log(carYears);
+console.log(" ");
 
 // ==== Challenge 5 ====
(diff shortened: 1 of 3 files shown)


assignments/arrays.js
```diff
@@ -63,34 +63,54 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+console.log(`Car 33 is a ${inventory[32].car_year} ${inventory[32].car_make} ${inventory[32].car_model}` );
 
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
 let lastCar = 0;
```
enitchals, line 72: They were expecting you to set `lastCar` to `inventory[inventory.length-1]`, but the way you did it is fine, too. I did it the same way you did, and I think that writing out the whole chain is a good way to reinforce the underlying structure. That said, in production-level code there are a lot of good reasons to use variables to represent values you're going to use repeatedly.

assignments/arrays.js
```diff
@@ -63,34 +63,54 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 // ==== Challenge 5 ====
 // The car lot manager needs to find out how many cars are older than the year 2000. Using the carYears array you just created, find out how many cars were made before the year 2000 by populating the array oldCars and logging it's length.
 let oldCars =[];
-console.log(); 
+for (let i = 0; i < carYears.length; i++) {
+    if (carYears[i] < 2000) {
+        oldCars.push(carYears[i]);
+    }
+}
+console.log(oldCars);
```
enitchals, line 102: Close -- your console.log statement should be `oldCars.length`, since they were looking for a number this time rather than a list.

assignments/callbacks.js
```diff
@@ -1,34 +1,74 @@
...
-function firstItem(arr, cb) {
+const cb = (x) => {console.log(x)};
+
+/* function firstItem(arr, cb) {
   // firstItem passes the first item of the given array to the callback function.
-}
+  return cb(arr[0]);
+} */
+
+let firstItem = (arr, cb) => cb(arr[0])
```
enitchals, line 10: One-line arrow functions! :+1:

assignments/callbacks.js
```diff
@@ -1,34 +1,74 @@
...
+  }
+  cb(condition);
+}    */
+
+ let contains = (item, items, cb) => cb(items.includes(item) ? true : false)
+
 
 /* STRETCH PROBLEM */
 
 function removeDuplicates(array, cb) {
```
enitchals, line 61: This was a great attempt (and it feels familiar, like I've tried a similar approach to a similar problem), but when I run the code, it actually _generates_ duplicates rather than removing them. I recommend looping through the array, looking at whether the `finalArray` already contains the element at `array[i]` and, if not, pushing `array[i]` to the `finalArray` list.

assignments/objects.js
```diff
@@ -19,26 +19,73 @@ let example = {
...
+  "email": "mmelloy0@psu.edu",
+  "gender": "F"
+}
+
+let intern2 = {
+  "id": 2,
+  "name": "Kennan",
+  "email": "kdiben1@tinypic.com",
+  "gender": "M",
+  speak: function() {
```
enitchals, line 34: Nice work adding the method directly to the object!

assignments/objects.js
```diff
@@ -49,16 +96,44 @@ let example = {
...
+    return `Hello I am ${this.Name}.`;
+  },
+
+  "child": {
+    "Name": "George",
+    "Age": 50,
+    "speak": function() {
+      return `Hello I am ${this.Name}.`;
+    }
+  },
```
enitchals, line 112: Grandchild should be nested on the child object. As written, you have both the child and grandchild objects nested within the parent object.

enitchals (commented):
Nice job with this! I had a few minor suggestions, but overall I can tell you have a good understanding of JS fundamentals :smile:

PR #26, opened 2018-04-16
Wladimir Fraga

wladimir917:
## JS1

- [x] objects.js
- [x] arrays.js
- [x] callbacks.js

- [x] Stretch Task

## Good Habits
 - [x] Pull Request title is your first and last name (add any notes in the comment field)
-  [x] Code has been committed regularly
- [x] Commit comments are informative
- [x] Code is organized in a logical way

### DIFF
--- a/assignments/arrays.js
+++ b/assignments/arrays.js
@@ -2,7 +2,12 @@
 
 // The car dealer has all of their inventory housed in the array seen below.  Scroll down past the data to find out how you can help the car dealer.
 
-let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year":2009},
+let inventory = [
+    {"id":1,
+    "car_make":"Lincoln",
+    "car_model":"Navigator",
+    "car_year":2009},
+
 {"id":2,"car_make":"Mazda","car_model":"Miata MX-5","car_year":2001},
 {"id":3,"car_make":"Land Rover","car_model":"Defender Ice Edition","car_year":2010},
 {"id":4,"car_make":"Honda","car_model":"Accord","car_year":1983},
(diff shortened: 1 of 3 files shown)


assignments/arrays.js
```diff
@@ -61,36 +66,219 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
-
 
+console.log("==== Challenge 1 ==== \n");
+for(let i = 0; i < inventory.length; i++) {
+    if(inventory[i].id === 33) {
+        console.log(`Car 33 is a ${inventory[i].car_year} ${inventory[i].car_make} ${inventory[i].car_model}\n` );
+    }
+}
```
nphillips78, line 83: I like your if statement here! But remember the array index starts at `0` so you'd actually be looking for index `32`.

assignments/arrays.js
```diff
@@ -61,36 +66,219 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+
+console.log("==== Challenge 5 ==== \n");
 let oldCars =[];
-console.log(); 
+
+for(let i = 0; i < carYears.length; i++) {
+    if(carYears[i] <= 2000) {
+        oldCars.push(carYears[i].car_year);
+    }
+}
```
nphillips78, line 132: What happens if the car is newer than 2000?

assignments/callbacks.js
```diff
@@ -1,34 +1,88 @@
...
+/*
+$ node assignments/callbacks.js
+First item in Items is : Pencil
+Length of Items is : 4
+Last item in Items is : Gum
+Sum numbers : 11
+Multiply numbers : 30
+Items Contains Pencil : true
+Items Contains Car: false
+*/
```
nphillips78, line 88: Good job!

assignments/objects.js
```diff
@@ -19,6 +19,48 @@ let example = {
...
+  "id": 5,
+  "name": "Antonietta",
+  "email": "adaine5@samsung.com",
+  "gender": "F",
+  "multiplyNums": function (a, b) {
+    return a * b;
+  }
+}
+
+
```
nphillips78, line 63: Great objects

assignments/objects.js
```diff
@@ -19,6 +19,48 @@ let example = {
...
+}
+
+let antonietta = {
+  "id": 5,
+  "name": "Antonietta",
+  "email": "adaine5@samsung.com",
+  "gender": "F",
+  "multiplyNums": function (a, b) {
+    return a * b;
+  }
```
nphillips78, line 60: Looks like you have some stray code here.

assignments/objects.js
```diff
@@ -49,16 +107,73 @@ let example = {
...
+==== Stretch Challenge: Nested Objects and the this keyword ====
+
+Parent's name Susan
+Child's age 50
+Grandchild's name Sam and age 30
+Parent can speak now!
+Child can speak now!
+Grandchild can speak now!
+
+*/
```
nphillips78, line 179: Good job

nphillips78 (commented):
Great work Vlad!

assignments/arrays.js
wladimir917: I know, but if this array comes from a database or the array is in a different order or you delete one car before, then not always is going to be in the position #32.

assignments/arrays.js
wladimir917: "The car lot manager needs to find out how many cars are older than the year 2000"

Sorry I skipped the first part, I will fix it

assignments/arrays.js
nphillips78: In this problem, you're given an array and asked for the 33rd car. The 33rd car, no matter how you mess with the array, will be at index 32.

If it were drawn from a database, you're absolutely right. We typically try to assign ids that are independent of the position in the database to avoid exactly this confusion. (for example, car id #3a could be moved around to different spots on the list)

EXCELLENT thinking on this, Vlad!

nphillips78: And on re-reading the problem, it looks like there was some ambiguity in the instructions.  So good job! Do you see how the use cases differ?

PR #35, opened 2018-04-16
Completed 

ghost:


assignments/arrays.js
```diff
@@ -63,34 +63,48 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+console.log(`Car 33 is a ${inventory[32].car_year} ${inventory[32].car_make} ${inventory[32].car_model}` );
```
designerexpert, line 66: Great way to do this without a loop, direct and simple 👍

assignments/arrays.js
```diff
@@ -63,34 +63,48 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
 let lastCar = 0;
-console.log();
+console.log(`${inventory[inventory.length - 1].car_make} ${inventory[inventory.length - 1].car_model}`);
```
designerexpert, line 73: good use of the length property to find the last item 💯 
Consider using descriptive console.logs like so: 
```console.log(`The last car in the inventory is a: ${inventory[inventory.length - 1].car_make} ${inventory[inventory.length - 1].car_model}`);```

File: assignments/arrays.js [hunk 3]
```diff
@@ -63,34 +63,48 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
 let carModels = [];
-console.log();
+for(let i = 0; i < inventory.length; i++){
+    carModels.push(inventory[i].car_model);
+}
+carModels = carModels.sort();
```
designerexpert, line 81: Good way to sort, outside the loop.

assignments/arrays.js
```diff
@@ -63,34 +63,48 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 // The car lot manager needs to find out how many cars are older than the year 2000. Using the carYears array you just created, find out how many cars were made before the year 2000 by populating the array oldCars and logging it's length.
 let oldCars =[];
-console.log(); 
+for(let i = 0; i < carYears.length; i++){
+    if(carYears[i] < 2000){
+        oldCars.push(carYears[i]);
+    }
+}
+
+console.log(oldCars.length);
```
designerexpert, line 103: Here again being more descriptive in our console log, can help us identify the text being shown in the console and differentiate it from all other outputs. I would do it like so: 
`console.log('There are ',oldCars.length, ' cars older than year 2000'); `

assignments/callbacks.js
```diff
@@ -2,27 +2,38 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 
 function multiplyNums(x, y, cb) {
   // multiplyNums multiplies two numbers and passes the result to the callback.
+  cb(x * y);
 }
 
 function contains(item, list, cb) {
   // contains checks if an item is present inside of the given array/list.
   // Pass true to the callback if it is, otherwise pass false.
+  if(list.include(item)) {
```
designerexpert, line 31: Great use of include array prototype.

assignments/callbacks.js
```diff
@@ -31,4 +42,15 @@ function removeDuplicates(array, cb) {
...
+  let arrFreeArry = [];
+  for(let i = 0; i < array.length; i++){
+    if(lastItem !== array[i]){
+      arrFreeArry.push(array[i]);
+      lastItem = array[i];
+    } else {
+      lastItem = array[i];
+    }
+  }
+  cb(arrFreeArr);
```
designerexpert, line 55: **ERROR:** `cb(arrFreeArr);` should be instead: `cb(arrFreeArry);`  Missing a 'y'.
Also, this function is not removing any duplicates from the array. It instead checks to see if the last item is the same as the one it is currently at. When checked against an array like so: 
```const items = ['Pencil', 'Gum','Notebook', 'yo-yo', 'Gum'];``` 
It fails to detect the duplicate item.
This is how I tested it: ```removeDuplicates(items, console.log)```
Perhaps utilizing includes could be beneficial in here. Also Filter would be very useful in this situation.

File: assignments/objects.js [hunk 7]
```diff
@@ -18,25 +18,56 @@ let example = {
...
-
+console.log(intern4.name)
 // Antonietta's Gender
-
+console.log(intern1.gender)
 // ==== Challenge 3: Object Methods ==== 
 // Give Kennan the ability to say "Hello, my name is Kennan!" Use the console.log provided as a hint.
 // console.log(kennan.speak());
-
+console.log(intern2.speak)
```
designerexpert, line 70: The speak was intended to be a function added to the object. Like so:
```intern2.speak = function() { return 'Hello, my name is '+ this.name + '!'; }```
```console.log(intern2.speak()); // Invoking that function like so.```
Try using this same pattern to adjust every other object's speak method.

PR #37, opened 2018-04-16
Erinc Emer JavaScript-I Completed

erinc35:


assignments/arrays.js
```diff
@@ -63,34 +63,52 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+console.log('Car 33 is a ' + inventory[32].car_year + ' ' + +inventory[32].car_make + ' ' + inventory[32].car_model);
```
enitchals, line 66: When I run this code, I get `Car 33 is a 2011 NaN Wrangler.` It looks like the second `+` before `inventory[32].car_make` is what's causing that -- when I remove it, I get the right solution.
What you have generally works as a way to piece together a string, but I'd typically recommend using a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) instead.

assignments/arrays.js
```diff
@@ -63,34 +63,52 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
-let lastCar = 0;
-console.log();
+let lastCar = inventory[inventory.length-1];
+console.log(lastCar);
```
enitchals, line 73: You did output the make and model, but also some additional information. Make sure you're isolating the pieces that were asked for.

assignments/callbacks.js
```diff
@@ -31,4 +37,8 @@ function removeDuplicates(array, cb) {
   // removeDuplicates removes all duplicate values from the given array.
   // Pass the duplicate free array to the callback function.
   // Do not mutate the original array.
+  cb(arr.reduce((dupless, item) => {
```
enitchals, line 40: Great solution to this stretch problem!

assignments/objects.js
```diff
@@ -49,16 +90,39 @@ let example = {
...
+let parent = {
+  'name': 'Susan',
+  'age': 70,
+  'child': {
+    'name': 'George',
+    'age': 50,
+    'grandchild': {
+      'name': 'Sam',
+      'age': 30,
+      speak: function () {
```
enitchals, line 102: I like how you organized this so that the methods were included _after_ the nested objects. It looks nice and it helps to visually break up the cascade of closing brackets.

enitchals (commented):
Great work! There are a couple minor errors that held you back, but I can tell you understand what's going on :+1:

PR #40, opened 2018-04-16
Ja Navey - JSI

JSNavey:
JSI - Objects, Arrays and Callbacks

### DIFF
--- a/assignments/callbacks.js
+++ b/assignments/callbacks.js
@@ -2,27 +2,33 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
 
 function firstItem(arr, cb) {
   // firstItem passes the first item of the given array to the callback function.
+  cb(arr[0]);
 }
 
 function getLength(arr, cb) {
   // getLength passes the length of the array into the callback.
+  cb(arr.length);
 }
 
 function last(arr, cb) {
   // last passes the last item of the array into the callback.
+  cb(arr[arr.length - 1]);
 }
 
 function sumNums(x, y, cb) {
   // sumNums adds two numbers (x, y) and passes the result to the callback.
+  cb(x + y);
 }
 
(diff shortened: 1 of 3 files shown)


assignments/callbacks.js
```diff
@@ -31,4 +37,10 @@ function removeDuplicates(array, cb) {
   // removeDuplicates removes all duplicate values from the given array.
   // Pass the duplicate free array to the callback function.
   // Do not mutate the original array.
-}
+  let duplicateFree = [];
```
enitchals, line 40: Nice job with this stretch problem! I think this is actually a better solution than the model.

assignments/objects.js
```diff
@@ -15,31 +15,89 @@ let example = {
...
+
 // Antonietta's Gender
 
+console.log(internFive.gender);
+
 // ==== Challenge 3: Object Methods ==== 
 // Give Kennan the ability to say "Hello, my name is Kennan!" Use the console.log provided as a hint.
 // console.log(kennan.speak());
 
+internTwo.speak = function() {
```
enitchals, line 85: Nice work! You can also add a method directly to the object. We'll talk more about methods when we get into Classes later this week :smile:

assignments/objects.js
```diff
@@ -49,16 +107,52 @@ let example = {
 // 3. Nest a grandchild object in the child object with properties for name and age.  The name will be Sam and the age will be 30
 // 4. Give each of the objects the ability to speak their names using the this keyword.
 
-let parent = {}
+let parent = {
```
enitchals, line 110: Great work with the nested object structure! It looks like you've got it down :+1:

enitchals (commented):
Great work! Please let me know if you want to talk more about any of these concepts, but it looks like you're developing a solid JavaScript foundation :)

PR #43, opened 2018-04-16
Daily PR Jorge Valdivia

ghost:


assignments/arrays.js
```diff
@@ -63,19 +63,35 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
-
-
+// for (let i = 0; i < inventory.length; i++) {
+//     if (inventory[i].id === 33) {
+//         console.log(inventory[i].car_make);
```
designerexpert, line 68: Console logs can be chained after each other like so: `console.log('first', 'second', 'third');` this saves lots of time.
Consider using descriptive console.logs to be able to identify where in the logs this is happening, and to better view your console outputs. I do this like so:
```console.log(`Car 33 is a: ${inventory[i].car_make} ${inventory[i].car_model} ${inventory[i].car_year}`)```

File: assignments/arrays.js [hunk 2]
```diff
@@ -63,19 +63,35 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
-
-
+// for (let i = 0; i < inventory.length; i++) {
```
designerexpert, line 66: FYI: This code is all commented out.

assignments/arrays.js
```diff
@@ -63,19 +63,35 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
-console.log();
+var unsorted = [];
+var carModels = [];
+for (let i = 0; i < inventory.length; i++) {
+    unsorted = inventory[i].car_model;
+    // unsorted[i].sort;
+    console.log(unsorted);
+    // console.log(unsorted);
+}
+
```
designerexpert, line 94: Adding a console.log in this line to list your unsorted array is better than logging it every time 'i' is incremented. This way your console does not get cluttered. 
Alternatively using the array.sort method might be useful to sort said array.

assignments/objects.js
```diff
@@ -18,27 +18,71 @@ let example = {
...
+  "gender": "F"
+}
 
+let internTwo = {
+  "id": 2,
+  "name": "Kennan",
+  "email": "kdiben1@tinypic.com",
+  "gender": "M",
+  "speak": function() {
+    console.log("Hello, my name is Kennan!");
```
designerexpert, line 34: Consider using this.name instead of Keenan inside the console.log to see a very interesting result. Like so:
```console.log("Hello, my name is " + this.name + "!");```
Then you can use the same pattern to approach all other objects in the future, the keyword `this` refers to the object in which the property is contained. in this case this refers to internTwo.

PR #61, opened 2018-05-21
Terrance Danckert CS11 (JavaScript-I)

Tbarnett20:


assignments/arrays.js
```diff
@@ -75,6 +75,13 @@ let inventory = [
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
+
+for (i = 0; i < inventory[i].length; i++) {
+  if (inventory[i].id === 33) {
+    console.log(`Car 33 is ${inventory[i].car_make} ${inventory.car_model}`);
+  }
+}
+
```
brcampbell001: When I run this code, it does not give me any output. Try to simplify by just using string template and calling that specific vehicle information instead of doing for loop

brcampbell001 (commented):
Great job on everything Terrance! The only problems that I would revisit are Challenge 1, 3, and 6.  They are not putting out the correct logs. They are close, just need a little reworking.

assignments/arrays.js
Tbarnett20: I left some text in my console log that didn't need to be there. I'll fix it

Tbarnett20: Lol okay I see what you mean. That was my fault trying to go fast.

PR #77, opened 2018-05-21
Libby Hart - cs11 [JavaScript-I]

LibbyHart:


assignments/objects.js
```diff
@@ -57,21 +57,34 @@ let antonietta = {
...
 
 // Antonietta loves math, give her the ability to multiply two numbers together and return the product. Use the console.log provided as a hint.
-//console.log(antonietta.multiplyNums(3,4));
+
+antonietta.multiplyNums = function(a,b) {
+  return a * b;
+}
+console.log(antonietta.multiplyNums(3,4));
 
 // === Great work! === Head over to the the arrays.js file or take a look at the stretch challenge
```
brcampbell001, line 89: Great work on objects! Try at the stretch problem when you get a chance. You are doing great!

assignments/arrays.js
```diff
@@ -72,25 +78,48 @@ console.log(`Car 33 is a *car year goes here* *car make goes here* *car model go
 let lastCar = 0;
 console.log();
 
+let lastCar = inventory[inventory.length - 1];
```
brcampbell001: Remember that a let variable cannot be declared twice so you would just want lastCar here without adding another let to it.

assignments/arrays.js
```diff
@@ -72,25 +78,48 @@ console.log(`Car 33 is a *car year goes here* *car make goes here* *car model go
...
+console.log('${lastCar.car_make} ${lastCar.car_model}');
```
brcampbell001: Remember with a template string you will need to use the back tick ( ` ) instead of the single quotation ( ' ). Check your console.log of this problem, it is printing out ${lastCar.car_make} ${lastCar.car_model}

assignments/arrays.js
```diff
@@ -63,7 +63,13 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+// console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+
+for(let i = 0; i < inventory.length; i++) {
+    if (inventory[i].id === 33) {
+        console.log('Car 33 is a ${inventory[i].car_year} ${inventory[i].car_make} ${inventory[i].car_model}');
```
brcampbell001: Need a template string here as well. Always remember to use back ticks ` (the key left to your 1 on the keyboard)

brcampbell001 (commented):
Great job so far! Your objects are excellent. There were a few things in the arrays exercise, but they are common errors to make. Always make sure that template strings are using back ticks instead of quotes (I struggled with this for way too long when I was doing this stuff). All in all, impressive work!

PR #130, opened 2018-06-25
Katherine Gilligan

katiagilligan888:


assignments/arrays.js
```diff
@@ -63,34 +63,56 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+console.log("Car 33 is a " + inventory[32]['car_year']  + " " + inventory[32]['car_make'] + " " + inventory[32]['car_model']);
```
dtacheny, line 66: This is good! I would recommend using dot notation whenever it's possible, but this is functionally the same.

assignments/callbacks.js
```diff
@@ -2,27 +2,49 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
 
 function firstItem(arr, cb) {
   // firstItem passes the first item of the given array to the callback function.
+  if(typeOf === "function"){
```
dtacheny, line 5: I like that you thought to check to make sure that it's a function. I think you need it to be `typeof cb === 'function'` to check the cb.

dtacheny (commented):
Great work 👍

PR #132, opened 2018-06-25
Rachel DiCesare

RachelDiCesare93:
first commit of the day


groov1234:
Hey, Rachel. I was able to test your Arrays assignment. The logic on the more difficult problems was solid.
`console.log()` statements for the first part were however, incorrect. ```console.log(`Car 33 is a 2011 Jeep Wrangler` );``` It would be too easy to just have a look at the array. Try logging your `inventory[i].id` on problem one and see what you return. Then I would give this a read for how to add the rest of the `console.log()` statement. https://developer.mozilla.org/en-US/docs/Web/API/console

Overall, good work. Need to work on syntax a bit, but your algorithms were sound.

PR #135, opened 2018-06-25
Mike Kerbleski

MKerbleski:


assignments/callbacks.js
```diff
@@ -1,34 +1,74 @@
...
+return cb(x*y);
 }
 
+console.log("\nmultiplyNums Answer: ")
+console.log(multiplyNums(4,4, test));
+
 function contains(item, list, cb) {
   // contains checks if an item is present inside of the given array/list.
   // Pass true to the callback if it is, otherwise pass false.
+  let variable = null;
```
dtacheny, line 52: This whole function is logically correct, but `variable` will never be `false`. I would initialize it to `false`, so then if `list` doesn't include `item`, it will be passed into the callback as `false` and not `null`.

assignments/objects.js
```diff
@@ -17,32 +17,77 @@ let example = {
...
 // Antonietta's Gender
+console.log(antonietta.gender);
 
-// ==== Challenge 3: Object Methods ==== 
+// ==== Challenge 3: Object Methods ====
 // Give Kennan the ability to say "Hello, my name is Kennan!" Use the console.log provided as a hint.
-// console.log(kennan.speak());
+
+
+function speak(object){
```
dtacheny, line 73: So here we want to give the `kennan` object a method called `speak`. You can do this just like you would assign any value to a variable, only instead of a string or number, it will be a function.

```
kennan.speak = function speak() {
    return 'Hello my name is Kennan.';
}
```

I really like how you thought about this though, passing the object in as an argument and then pulling the name from the object!

assignments/objects.js
```diff
@@ -17,32 +17,77 @@ let example = {
...
+  return ('Hello, my name is ' + object['name']);
+}
+
+console.log(speak(kennan));
 
 // Antonietta loves math, give her the ability to multiply two numbers together and return the product. Use the console.log provided as a hint.
-//console.log(antonietta.multiplyNums(3,4));
+
+
+function multiplyNums(x,y){
```
dtacheny, line 82: Same as above! Only we want to add `multiplyNums` to `antonietta`!

dtacheny (commented):
Overall great job 👍

PR #138, opened 2018-06-25
David White - Objects Completed

kauaicrusier:


assignments/arrays.js
```diff
@@ -63,14 +63,21 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+car33 = []
+for (i = 0; inventory.length; i++) {
+  if ("id" = 33) {
```
dtacheny, line 68: We want to pull the `id` off of the element in `inventory` at the current index. So since we're iterating over the `inventory` array, we need to look at an individual element and _then_ pull the id off.

We can do this by first accessing the individual element:
```
inventory[i]
```

and since `inventory[i]` is an object (which has a property of `id`), we then use dot notation to get the id.
```
inventory[i].id
```

This is what you want to use in the conditional on line 68. Right now, that line is comparing the string `"id"` with the number `33`, which will always fail.

assignments/arrays.js
```diff
@@ -63,14 +63,21 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+  }
+}
 
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
 let lastCar = 0;
-console.log();
+  lastCar = inventory.length;
```
dtacheny, line 78: `inventory.length` is going to give you the number of elements inside of inventory, which is going to be 50.

This is a problem. Because JavaScript uses zero-based indexing (the first element has an index of 0), the array does _not_ have an element at the 50th index. The last element in the array is going to have an index of 49. So trying to access `inventory[50]` will cause an error.

So two things need to happen here:

1. We need to get the correct index for the last car in the array.
1. We need to assign `lastCar` to the last element in the list.

For the first one, we have to subtract one from the length of the array. So the index of the last element in `inventory` is going to be:
```
inventory.length - 1
```

So for number two, how do we get the last element in the array? Using the index above:
```
inventory[inventory.length - 1]
```
This will give us the last object in the `inventory` array, which we can then assign to `lastCar`. _Then_ you can use the dot notation in the console logs and get the correct output, since `lastCar` now refers to an object with `car_make` and `car_model` properties.

assignments/arrays.js
```diff
@@ -80,17 +87,29 @@ console.log();
 // ==== Challenge 4 ====
 // The accounting team needs all the years from every car on the lot. Create a new array from the dealer data containing only the car years and log the result in the console.
 let carYears = [];
+  for(let i = 0; i < inventory.length; i++) {
+    years = inventory.car_year;
```
dtacheny, line 91: `inventory` is an array, so it doesn't have a `.car_year` property. You need to get `car_year` off of each individual element within `inventory`. So all you have to do here is change line 91 to refer to the element at an index in the array instead of referring to the array itself:
```
years = inventory[i].car_year;
```
But you could also combine lines 91 and 92!
```
carYears.push(inventory[i].car_year);
```

assignments/arrays.js
```diff
@@ -80,17 +87,29 @@ console.log();
...
+    carYears.push(years);
+  }
 console.log();
 
 // ==== Challenge 5 ====
 // The car lot manager needs to find out how many cars are older than the year 2000. Using the carYears array you just created, find out how many cars were made before the year 2000 by populating the array oldCars and logging it's length.
 let oldCars =[];
-console.log(); 
+  for (let i = 0; i < carYears.length; i++) {
+    if (car_year < 2000) {
```
dtacheny, line 100: This is very close! `car_year` should be `carYears[i]` since we are iterating over the carYears, and we want to compare each element to `2000`. Right now, car_year doesn't refer to anything.

assignments/arrays.js
```diff
@@ -80,17 +87,29 @@ console.log();
...
+console.log();
 
 // ==== Challenge 6 ====
 // A buyer is interested in seeing only BMW and Audi cars within the inventory.  Return an array that only contains BMW and Audi cars.  Once you have populated the BMWAndAudi array, use JSON.stringify() to show the results of the array in the console.
 let BMWAndAudi =[];
-console.log();
-
-
+  for(i = 0; inventory.length; i++)
+    if(car_make === BMW || Audi) {
```
dtacheny, line 111: `car_make` doesn't refer to anything at this point. We need to get it from the element in the array.
```
inventory[i].car_make === 'BMW'
```
Remember to make `BMW` a string.

And another annoying thing about `||` and `&&` operators is that you have to type out the whole thing again. So you would need the condition to be
```
inventory[i].car_make === 'BMW' || inventory[i].car_make === 'Audi'
```

assignments/objects.js
```diff
@@ -18,31 +18,78 @@ let example = {
...
 // Give Kennan the ability to say "Hello, my name is Kennan!" Use the console.log provided as a hint.
 // console.log(kennan.speak());
 
+  function 
+
+
+
 // Antonietta loves math, give her the ability to multiply two numbers together and return the product. Use the console.log provided as a hint.
 //console.log(antonietta.multiplyNums(3,4));
+  function antonietta (x, y) {
```
dtacheny, line 84: So here, we want to add a method to the object that refers to Antonietta, in this case you have her as `student5`.

Sort of like how `student5` has a `gender` property, we can add functions in the same way. Here, we want to add a function to `student5`

This can be done just like assigning something to a variable:
```
student5.multiplyNums = function(x, y) {
   return x * y;
}
```

This way you can call `student5.multiplyNums(4, 6)` just like any other function.

dtacheny (commented):
Seems like you have a lot of the logic down, but `for` loops are kind of confusing. Let me know if I can help to clarify how those work!

PR #142, opened 2018-06-25
Alberto Preciado: Stretch Complete

albertopfunk:


assignments/callbacks.js
```diff
@@ -2,33 +2,110 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+/*
+const last = (arr, cb) => cb(arr[arr.length - 1]);
+*/
+
+
+
 function sumNums(x, y, cb) {
   // sumNums adds two numbers (x, y) and passes the result to the callback.
+  let result = x + y;
+  cb(result);
```
EricHech: `cb(x + y)` should work as well. Same for the multiplication one below.

assignments/callbacks.js
```diff
@@ -2,33 +2,110 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 }
+*/
+
+
 
 function contains(item, list, cb) {
   // contains checks if an item is present inside of the given array/list.
   // Pass true to the callback if it is, otherwise pass false.
+  if (list.includes(item)) {
```
EricHech, line 68: These could have all been one-lined, but many linters require it to be written as you did.

`if (list.includes(item)) cb(true);`
`else cb(false);`

assignments/callbacks.js
```diff
@@ -2,33 +2,110 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 
+/*
+const contains = (item, list, cb) => list.includes(item) ? cb(true) : cb(false);
+*/
+
+
+
 /* STRETCH PROBLEM */
 
 function removeDuplicates(array, cb) {
```
EricHech, line 83: Nice job.

assignments/objects.js
```diff
@@ -49,16 +96,42 @@ let example = {
...
+  name: 'Susan',
+  age: 70,
+  child: {
+    name: 'George',
+    age: 50,
+    grandchild: {
+      name: 'Sam',
+      age: 30,
+      speak: function() {
+        return `Hello my name is ${this.name}`
```
EricHech: So I'm sure you saw from our standup last night how it would perhaps be better to declare the speak function outside of the object and then assign it in each object as a method so that you don't have to write the same function multiple times. If you do this, remember not to call (`speak()`) the function when you set it on the key.

EricHech (commented):
Nice work.

assignments/objects.js
albertopfunk: Ya I was trying to do that but it wasn't working so I just did it the way I did to complete the project, i'll look into it and refactor

PR #144, opened 2018-06-25
finished objects..starting arrays

rverdi642:
Richard Verdier

### DIFF
--- a/assignments/arrays.js
+++ b/assignments/arrays.js
@@ -70,27 +70,55 @@ console.log(`Car 33 is a *car year goes here* *car make goes here* *car model go
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
 let lastCar = 0;
-console.log();
+let inventoryLength = inventory.length;
+console.log()
+console.log(inventoryLength);
+console.log(("Make: " +inventory[ inventoryLength-1].car_make + " Model: " +inventory[ inventoryLength-1].car_model));
 
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
 let carModels = [];
-console.log();
+
+let inventoryLength1 = inventory.length;
+
+for (i=0; i< inventoryLength1;i++){
+
+ carModels[i] = inventory[i].car_model
+//  console.log(carModels[i])
+ }
+
+console.log(carModels.sort());
 
 // ==== Challenge 4 ====
 // The accounting team needs all the years from every car on the lot. Create a new array from the dealer data containing only the car years and log the result in the console.
 let carYears = [];
-console.log();
+
+for (i=0; i< inventory.length;i++) {
+
+ carYears[i] = inventory[i].car_year;
+ 
+}
+
+ console.log(carYears);
+
 
 // ==== Challenge 5 ====
 // The car lot manager needs to find out how many cars are older than the year 2000. Using the carYears array you just created, find out how many cars were made before the year 2000 by populating the array oldCars and logging it's length.
-let oldCars =[];
-console.log(); 
+ let oldCars =[];
+ for (i=0; i< inventory.length;i++) {
+   if(parseInt(inventory[i].car_year) < 2000){
+  oldCars[i]=parseInt(inventory[i].car_year);
+  //  console.log(oldCars.length); 
+  }
+
+}
+
+console.log(oldCars); 
 
 // ==== Challenge 6 ====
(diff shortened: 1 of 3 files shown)


assignments/arrays.js
```diff
@@ -70,27 +70,55 @@ console.log(`Car 33 is a *car year goes here* *car make goes here* *car model go
...
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
 let carModels = [];
-console.log();
+
+let inventoryLength1 = inventory.length;
+
+for (i=0; i< inventoryLength1;i++){
+
+ carModels[i] = inventory[i].car_model
```
dtacheny, line 86: Line 86: This works, however I would recommend using `carModels.push()` to add the elements. This way is fine for this problem because you're doing it for every element in `inventory`, but it will leave holes in an array (the element at some indexes would be `undefined`) if you were to only add some elements.

assignments/arrays.js
```diff
@@ -70,27 +70,55 @@ console.log(`Car 33 is a *car year goes here* *car make goes here* *car model go
...
+
 
 // ==== Challenge 5 ====
 // The car lot manager needs to find out how many cars are older than the year 2000. Using the carYears array you just created, find out how many cars were made before the year 2000 by populating the array oldCars and logging it's length.
-let oldCars =[];
-console.log(); 
+ let oldCars =[];
+ for (i=0; i< inventory.length;i++) {
+   if(parseInt(inventory[i].car_year) < 2000){
+  oldCars[i]=parseInt(inventory[i].car_year);
```
dtacheny, line 110: This is a situation where we have holes in our array. Because we are not adding every element in `inventory`, but the index, `i` is still incrementing after every iteration of the loop, we have a bunch of `undefined` elements in the `oldCars` array. If you log `oldCars` to the console, you get something like this:
```
[undefined, undefined, undefined, 1983, 1990, 1995, undefined, 1987, 1996, undefined]
```
If you use the `.push()` method, we don't get those `undefined`s since `.push` doesn't rely on the current index in `inventory` to add the element.

assignments/callbacks.js
```diff
@@ -2,29 +2,117 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+firstItem(items,printItem)
+
+
+
+
+
 function getLength(arr, cb) {
   // getLength passes the length of the array into the callback.
+  
+  const item = cb(arr);
```
dtacheny, line 23: We want to pass the length of `arr` into the callback. So for this, we would do:
```
const item = cb(arr.length);
```

assignments/callbacks.js
```diff
@@ -2,29 +2,117 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+firstItem(items,printLength)
+
+
+
+
 function last(arr, cb) {
-  // last passes the last item of the array into the callback.
+  // last passes the last item of the array into the callback
+  
+  const item =cb(arr);
```
dtacheny, line 40: Here we want to pass the last element in the array into the callback.
```
const item = cb(arr[arr.length - 1]);
```

assignments/callbacks.js
```diff
@@ -2,29 +2,117 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+
+last(items,printItem)
+
+
+
+
+
 function sumNums(x, y, cb) {
   // sumNums adds two numbers (x, y) and passes the result to the callback.
+  const sum = cb(x,y);
```
dtacheny, line 56: So, you wrote the callback function to do the adding, which shows you understand how the callback works with the higher-order function.

For this problem, we want to pass `x + y` into the callback.
```
const sum = cb(x + y);
```
This will be the same for the next few problems.

I appreciate what you did here and in the next few problems, though. 👏

assignments/callbacks.js
```diff
@@ -2,29 +2,117 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 
 function contains(item, list, cb) {
   // contains checks if an item is present inside of the given array/list.
   // Pass true to the callback if it is, otherwise pass false.
+  let vlaue = cb(item,list);
+  console.log(value);
+  
 }
 
+function contains(item, list, cb) {
```
dtacheny, line 89: For this one, your code gives the correct answer.

However, the implementation is a little inside out for what is being asked.

We want `contains` to be doing all of the work here, and to pass either `true` or `false` into the callback function, whatever the callback is.

So your code could be rearranged to fit the requirements by doing this:
```
  let i = list.indexOf(item);
  if(i < 0) {
    cb(false);
  }
  else {
    cb(true);
  }
}
```
I really like how you're thinking about these problems, I don't want you to think you're doing them wrong. These problems are really abstract and it's kind of hard to tell what you're supposed to be doing, so you're doing great.

dtacheny (commented):
👍

PR #147, opened 2018-06-25
Justin Hammett | JavaScript-I | In Progress 

hammettDev:


Nate152:
I know you were working on the objects challenges last I heard from you. Reach out to me sometime today and we can keep working on these problems together if need be. Think you got the array part down pretty quickly so its just a matter of time before it all clicks. Stick with it.

PR #148, opened 2018-06-25
Keith Kennedy 

Kennedykid1995:
I just started on the objects.js creating the objects for interns


assignments/arrays.js
```diff
@@ -63,14 +63,14 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+console.log(`Car 33 is a 2011 "jeep" "Wrangler"` );
```
TERR1E, line 66: The answer is technically correct but you were supposed to reference it using code.
1. Which code do you use to get a specific index value out of an array?
2. Which code do you use to get a specific value out of the property of an object?

The same goes for Challenge 2 below.

assignments/arrays.js
```diff
@@ -63,14 +63,14 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
 let lastCar = 0;
-console.log();
+console.log(`Last car is a "Lincoln" "Town Car"` )
```
TERR1E: Which code would you use to get the last index value of an array?

assignments/arrays.js
```diff
@@ -80,17 +80,19 @@ console.log();
 // ==== Challenge 4 ====
 // The accounting team needs all the years from every car on the lot. Create a new array from the dealer data containing only the car years and log the result in the console.
 let carYears = [];
+let carYears = [];
+for(let i=0; i < inventory.length; i++){
+  carYears.push(inventory[i].car_year);
+}
+console.log(carYears);
 console.log();
```
TERR1E: You don't need to put an empty console.log

assignments/objects.js
```diff
@@ -18,31 +18,66 @@ let example = {
...
+Kennan.speak = "Hello, my name is Kennan!";
 
+console.log(Kennan.speak)
 // Antonietta loves math, give her the ability to multiply two numbers together and return the product. Use the console.log provided as a hint.
 //console.log(antonietta.multiplyNums(3,4));
-
+function math(x, y){
+  return x*y;
+}
+console.log(math(3,4));
```
TERR1E, line 77: Check out the hint they've given to you:
`console.log(antonietta.multiplyNums(3,4));`

1. How would you name your function based on the hint?
2. What are the ways of assigning a new property to the existing object (Antonietta)?

Check out the links below:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_Objects

https://stackoverflow.com/questions/1184123/is-it-possible-to-add-dynamically-named-properties-to-javascript-object

http://www.dyn-web.com/tutorials/object-literal/properties.php

PR #150, opened 2018-06-25
Naaz Shaikh

noshaikh:


assignments/arrays.js
```diff
@@ -63,34 +63,69 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+
+console.log(inventory[32]);
```
Nate152, line 67: This will give us the data that we want but not in the proper syntax. We would rather access the individual properties on the objects and have it look a little cleaner. The easiest and cleanest way is with template literals. These are just a fancy syntax that allows us to access properties of JS objects within our console statements. For example:
```console.log(`Car 33 is a ${inventory[32].car_year} ${inventory[32].car_make} ${inventory[32].car_model}`)```
A couple of things to note about the syntax. Everything in the console log is in backticks (not single quotes) in order to allow us to access these properties on the object within a string. We then just use the ${} syntax with whatever property we want to reference inside the curly braces. Again these are called template literals and make it very easy to access properties later without having to assign them to variables.

File: assignments/arrays.js [hunk 2]
```diff
@@ -63,34 +63,69 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
-let lastCar = 0;
-console.log();
+let lastCar = (inventory.length)-1;
+console.log(inventory[lastCar]);
```
Nate152, line 74: Have to make sure we are correctly referencing the right place in the array. In line 73 you are setting lastCar equal to a number as inventory.length is going to return the number of items in the array. In order to access the last item we need to say ```inventory[inventory.length - 1]```. This will reference the last object in the inventory array. Once we assign the value of ```inventory[inventory.length -1]``` to ```lastCar``` we can reference it by simply calling lastCar. No need to continue using the bracket notation.

```
let lastCar = inventory[inventory.length - 1];
console.log(lastCar);
```

assignments/arrays.js
```diff
@@ -63,34 +63,69 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
-let carModels = [];
-console.log();
+let carModels=[]
+for (i=0; i<inventory.length; i++) {
+    
+    carModels.push((inventory[i]['car_model']));
```
Nate152, line 81: You could also have used dot notation here. Sometimes it can be a little more visually clear. Ex:
```carModels.push(inventory[i].car_model);```
Serves the exact same purpose and if brackets work for you then use them. I just wanted you to be aware of the different options. Dot notation is more popular.

File: assignments/arrays.js [hunk 4]
```diff
@@ -63,34 +63,69 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+ if ((carYears[i]) < 2000){
+   oldCars.push(carYears[i]);
+ }
+  console.log(oldCars)
+
+}
+
 
 // ==== Challenge 6 ====
 // A buyer is interested in seeing only BMW and Audi cars within the inventory.  Return an array that only contains BMW and Audi cars.  Once you have populated the BMWAndAudi array, use JSON.stringify() to show the results of the array in the console.
```
Nate152, line 114: Could have done this one without two loops. Your solution is definitely not wrong but it will be slower than:
```
let BMWAndAudi =[]; 
for (let i = 0; i < inventory.length; i++) {
  if (inventory[i].car_make == "BMW") {
    BMWAndAudi.push(inventory[i]);
  } else if(inventory[i].car_make == "Audi") {
    BMWAndAudi.push(inventory[i]);
  }
}
console.log("This is a list of all the BMW and Audi cars on the lot: " + JSON.stringify(BMWAndAudi));
```
or
```
let BMWAndAudi =[]; 
for (let i = 0; i < inventory.length; i++) {
  if (inventory[i].car_make === "BMW") || (inventory[i].car_make == "Audi") {
    BMWAndAudi.push(inventory[i]);
  }
}
console.log("This is a list of all the BMW and Audi cars on the lot: " + JSON.stringify(BMWAndAudi));
```
Note that the challenge wanted you to turn this into a JSON object using ```JSON.stringify```.

assignments/objects.js
```diff
@@ -18,28 +18,66 @@ let example = {
...
+"name" : "Antonietta",
+"gender" : "F",
+};
 
 // ==== Challenge 2: Reading Object Data ==== 
 // Once your objects are created, log out the following requests from HR into the console:
 
 // Mitzi's name
-
+console.log (Object.entries(intern1)[2]);
```
Nate152, line 60: You can access the property directly with ```console.log(intern1.name)```. No need for all of the extra object properties like entries. Different strokes for different folks. I will continue to say that yours is absolutely not wrong but sometimes there are shorter ways to write it and programmers are lazy by nature.

Nate152:
Looks good for day 1. Keep in mind that this was day 1 with JS. I think its important not to lose sight of that. Look through the comments and let me know if you have any questions.

PR #152, opened 2018-06-25
Jonathan Holloway

codejoncode:
completed arrays.js

### DIFF
--- a/assignments/arrays.js
+++ b/assignments/arrays.js
@@ -63,34 +63,59 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+for(car in inventory){
+  if(inventory[car]["id"] === 33){
+    console.log(`Car ${inventory[car]["id"]} is a ${inventory[car]['car_year']} ${inventory[car]['car_make']} ${inventory[car]['car_model']}`)
+  }
+}
+
 
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
-let lastCar = 0;
-console.log();
+let lastCar = inventory.length;
+for(car in inventory){
+  if(inventory[car]["id"] === lastCar){
+    console.log(`The last car is a ${inventory[car]['car_make']} ${inventory[car]['car_model']}`)
+  }
+}
 
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
 let carModels = [];
-console.log();
+for (car in inventory){
+  // console.log(inventory[car]["car_model"])
+  let model = inventory[car]["car_model"].toLowerCase(); 
+  carModels.push(model);
+}
+carModels = carModels.sort();
+console.log(carModels);
 
 // ==== Challenge 4 ====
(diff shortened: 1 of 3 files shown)


assignments/arrays.js
```diff
@@ -63,34 +63,59 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+for(car in inventory){
```
Nate152, line 66: Cool use of a for in loop. Did we need it though?
```console.log(`Car 33 is a ${inventory[32].car_year} ${inventory[32].car_make} ${inventory[32].car_model})```
This solution uses template literals which I don't think was covered in the lecture so if you don't necessary understand what this is doing shoot me a DM. They just allow us to directly access/reference properties within the object using a specific syntax.

File: assignments/arrays.js [hunk 2]
```diff
@@ -63,34 +63,59 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+
 
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
-let lastCar = 0;
-console.log();
+let lastCar = inventory.length;
+for(car in inventory){
```
Nate152, line 78: Similar to your first solution this definitely gets it done but you could have done it in a simpler way.
```
let lastCar = inventory[inventory.length -1];
console.log(lastCar);
```
No need to loop and create extra run time.

assignments/arrays.js
```diff
@@ -63,34 +63,59 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+  if(inventory[car]["id"] === lastCar){
+    console.log(`The last car is a ${inventory[car]['car_make']} ${inventory[car]['car_model']}`)
+  }
+}
 
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
 let carModels = [];
-console.log();
+for (car in inventory){
```
Nate152, line 87: Based on the repeated use of for in I am gonna assume that you know how to write a basic for loop however I would have liked you to have written it out once. For in has some use cases where it will not work and you are going to have to use a basic loop. Also am not sure it was necessary to use ```toLowerCase()```. Maybe you could convince me if we actually discussed it 😃

assignments/objects.js
```diff
@@ -50,15 +89,30 @@ let example = {
 // 4. Give each of the objects the ability to speak their names using the this keyword.
 
 let parent = {}
```
Nate152, line 91: We wanted to see you create a nested object here. Example:
```
let parent = {
  "name": "Susan",
  "age": 70,
  "speak": function() {
    return "My name is: " + this.name; 
  },
  "child": {
    "name": "George",
    "age": 50,
    "speak": function() {
      return "My name is: " + this.name; 
    },
    "grandchild": {
      "name": "Sam",
      "age": 30,
      "speak": function() {
        return "My name is: " + this.name; 
      }
    }
  }
}
```
And then based on this nesting structure:
```
// Log the parent object's name
console.log(parent.name);
// Log the child's age
console.log(parent.child.name)
// Log the name and age of the grandchild
console.log(parent.child.grandchild.name, parent.child.grandchild.age)
// Have the parent speak
console.log(parent.speak());
// Have the child speak
console.log(parent.child.speak());
// Have the grandchild speak
console.log(parent.child.grandchild.speak());
```

Nate152:
Check out my comments and make sure you don't have any questions. You are using some more advanced JS techniques so my assumption is that you understand the basics. JS can be approached in a ton of different ways and none of your solutions are wrong by any means but there are some specific concepts that were trying to have you show and thats what my comments pertain to.

PR #155, opened 2018-06-25
Rodrigo Acosta-Ramos

ghost:
Still working

### DIFF
--- a/assignments/arrays.js
+++ b/assignments/arrays.js
@@ -109,14 +109,20 @@ console.log(inventory.map(car=> car.car_year));
 // }
 console.log(inventory.map(car => car.car_year).filter(year => year < 2000)); 
 
+
 // ==== Challenge 6 ====
 // A buyer is interested in seeing only BMW and Audi cars within the inventory.  Return an array that only contains BMW and Audi cars.  Once you have populated the BMWAndAudi array, use JSON.stringify() to show the results of the array in the console.
-let BMWAndAudi =[];
-for (let i=0; i<len; i++) {
-  if ((inventory[i].car_make === 'BMW') || (inventory[i].car_make === 'Audi')) {
-    BMWAndAudi.push(inventory[i]);
-  }
+function germanLux(array) {
+  return JSON.stringify(inventory.filter(obj => (obj.car_make === 'BMW') || (obj.car_make === 'Audi')));
 }
-for (car of Object.entries(BMWAndAudi)) {
-  console.log(JSON.stringify(car));
-}
+console.log(germanLux(inventory));
+
+// let BMWAndAudi =[];
+// for (let i=0; i<len; i++) {
+//   if ((inventory[i].car_make === 'BMW') || (inventory[i].car_make === 'Audi')) {
+//     BMWAndAudi.push(inventory[i]);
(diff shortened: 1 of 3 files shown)


assignments/arrays.js
```diff
@@ -63,34 +63,66 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+
+let keys = Object.keys(inventory);
+let len = keys.length;
+// for (let i=0; i<len; i++) {
+//   console.log(inventory[i]);
+// }
+
+console.log('Car 33 is a ' + inventory[33].car_year.toString()
```
Nate152, line 73:
```
.toString()
```
This is not necessary. If we look at the data types, most of them are strings already and the console will print them without making them into strings. Very minor, we just don't need to be this explicit.

assignments/arrays.js
```diff
@@ -63,34 +63,66 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+            + ' ' + inventory[33].car_make.toString()
+            + ' ' + inventory[33].car_model.toString());
 
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
-let lastCar = 0;
-console.log();
+let lastCar = inventory[len-1];
```
Nate152, line 81: Took me a second to realize where you pulled the ```len``` from. You may wanna define it directly above the first time you use it so that its a little more clear. Another way we could have done with without defining that variable is:
``` 
let lastCar = inventory[inventory.length-1];
console.log(lastCar);
```

assignments/callbacks.js
```diff
@@ -1,34 +1,92 @@
...
+const items = ['Pencil', 'Gum', 'Notebook', 'yo-yo', 'Gum'];
+//===========================================
 function firstItem(arr, cb) {
   // firstItem passes the first item of the given array to the callback function.
+  return cb(arr);
+}
+function itemFirst(arr) {
+  return arr[0];
 }
+console.log(firstItem(items, itemFirst));
```
Nate152, line 10: I see what you are doing here, using the itemFirst as your cb function however there was a less convoluted way to handle this.
```
fn firstItem(arr, cb) {
cb(arr[0]);
}
````
This serves the same purpose and is a little more explicit that we are using a cb function. When we invoke the firstItem function, the first item in arr will be passed to the callback function. The wording of these questions makes it sound much more complicated than in actually is. This goes for all of the following problems. I'm fairly sure that you understand the concept, just want you to see that you could have written it all much simpler for this line and all the following instances where you plugged it into a console.log.

Nate152:
Overall a very good job. Love that you went back and refactored some of the array problems to use the built in array methods. You will use those over and over and over and over and over...... 😄 Glad to see that you are liking this better than previous topics.

PR #156, opened 2018-06-25
JoCatherine Griggs

jocafrin:
- [ ] Created intern objects

### DIFF
--- a/assignments/objects.js
+++ b/assignments/objects.js
@@ -19,6 +19,37 @@ let example = {
 
 // Write your intern objects here:
 
+let example = {
+  "id": 1,
+  "name": "Mitzi",
+  "email": "mmelloy0@psu.edu",
+  "gender": "F"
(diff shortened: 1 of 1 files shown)


DixieKorley:
Thanks for the PR JoCatherine! 👍

assignments/objects.js
```diff
@@ -19,6 +19,37 @@ let example = {
...
+  "email": "gmartinson3@illinois.edu",
+  "gender": "M"
+}
+let example = {
+  "id": 5 ,
+  "name": "Antonietta",
+  "email": "adaine5@samsung.com",
+  "gender": "F"
+}
+
```
TERR1E, line 52: Hey JoCatherine, you are on the right track here - except I would suggest giving specific names to your objects.
ex)
`let jocatherine = {}`
`let dixie = {}`

PR #162, opened 2018-06-25
Arthur Bates

AliasAJ:
working on array.js challenges

### DIFF
--- a/assignments/arrays.js
+++ b/assignments/arrays.js
@@ -2,56 +2,56 @@
 
 // The car dealer has all of their inventory housed in the array seen below.  Scroll down past the data to find out how you can help the car dealer.
 
-let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year":2009},
-{"id":2,"car_make":"Mazda","car_model":"Miata MX-5","car_year":2001},
-{"id":3,"car_make":"Land Rover","car_model":"Defender Ice Edition","car_year":2010},
-{"id":4,"car_make":"Honda","car_model":"Accord","car_year":1983},
-{"id":5,"car_make":"Mitsubishi","car_model":"Galant","car_year":1990},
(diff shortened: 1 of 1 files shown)


assignments/arrays.js
```diff
@@ -63,33 +63,46 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+console.log(`Car ${inventory[32].id} is a ${inventory[32].car_year} ${inventory[32].car_make} ${inventory[32].car_model}`);
 
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
-let lastCar = 0;
-console.log();
+let lastCar = inventory.length - 1;
+console.log(inventory[lastCar].car_make, inventory[lastCar].car_model)
```
TERR1E, line 73: After you assign the last item of inventory to `lastCar`, you do not need to define inventory again as it is stored to `lastCar` variable already.

assignments/arrays.js
```diff
@@ -63,33 +63,46 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
-let carModels = [];
-console.log();
+for (i = 0; i < inventory.length; i++) {
+    let carModels = inventory[i].car_model;
+    console.log(carModels);
+}
```
TERR1E, line 80: Your for loop above will not be storing all the car models; instead - it will constantly update `carModels` to the current iteration.
The empty array was given to you - which array method would you use to 'add' car models as each iteration goes?

The same goes for challenge 4, 5, and 6.

TERR1E (commented):
You almost got arrays.js - keep it up!

PR #163, opened 2018-06-25
Liz Baker

LizBaker:


assignments/arrays.js
```diff
@@ -63,34 +63,61 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 
 
 // ==== Challenge 2 ====
-// The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
-let lastCar = 0;
-console.log();
+// The dealer needs the information on the last car in their inventory.  
+// What is the make and model of the last car in the inventory? Log the make and model into the console.
+let lastCar = inventory[49];
```
Nate152, line 73: While this works, I think the idea that we wanted you to show here is something along the lines of:
```let lastCar = (inventory.length - 1)```
We wont always know the length of the array that we are trying to get data from so doing it this way is more explicit and will work every time. There could be cases were if something got pushed later into the inventory array, your version would no longer work.

File: assignments/callbacks.js [hunk 2]
```diff
@@ -2,27 +2,45 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 
+
 function getLength(arr, cb) {
   // getLength passes the length of the array into the callback.
+  cb(arr.length)
 }
 
 function last(arr, cb) {
   // last passes the last item of the array into the callback.
+  let lastArr = arr.length - 1
```
Nate152, line 16: Perfect example here of what I was talking about in the array assignment. This is much more explicit to give us that last item every time when we don't know the length of the array ahead of time.

assignments/objects.js
```diff
@@ -19,26 +19,77 @@ let example = {
 
 // Write your intern objects here:
 
+let interns = [
```
Nate152, line 22: Is there a particular reason that you decided to put these objects in an array? This solution is definitely valid and I like that you thought about it this way. Another way we could have gone about it is to just create the objects and ultimately I think thats what the assignment expected based on the example object. We don't have to have a container for them like if they were divs in html. The example had an individual object that was named and defined with a let. So for example:
```
let mitzi = {
id: 1,
name: "Mitzi",
email: "mmelloy0@psu.edu",
gender: "F"
}
```
We could then access each object later by simply referencing mitzi._insertKeyFromObjectHere_ . Also notice that I do not have quotes around the names of the keys. These are always going to be strings so we do not need to define them as such. The values can be anything so those we do need to define.

assignments/objects.js
```diff
@@ -19,26 +19,77 @@ let example = {
...
+  }
+}]
+
 
 // ==== Challenge 2: Reading Object Data ==== 
 // Once your objects are created, log out the following requests from HR into the console:
 
 // Mitzi's name
 
+console.log(interns[0].name);
```
Nate152, line 66: If we used the alternate way of creating the objects:
```
console.log(mitzi.name)
```

assignments/objects.js
```diff
@@ -19,26 +19,77 @@ let example = {
...
 // ==== Challenge 3: Object Methods ==== 
 // Give Kennan the ability to say "Hello, my name is Kennan!" Use the console.log provided as a hint.
-// console.log(kennan.speak());
+let kennan = interns[1];
+
+console.log(kennan.speak());
 
 // Antonietta loves math, give her the ability to multiply two numbers together and return the product. Use the console.log provided as a hint.
-//console.log(antonietta.multiplyNums(3,4));
+let antonietta = interns[4];
```
Nate152, line 91: Here you had to set antonietta because of how you chose to set it up earlier. Just a subtle difference and not anything wrong by any means. Different strokes for different folks.

Nate152:
Overall good job. Check out my comments and let me know if you have any questions or any concepts that I may be able to clear up.

PR #165, opened 2018-06-25
Steven Lai

ghost:
- [x] objective
- [x] stretch


assignments/arrays.js
```diff
@@ -61,36 +61,62 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 //     arr[i]; // 1,2,3,4
 // }
 
+
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
 console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
 
-
+for (let i = 0; i < inventory.length; i ++) {
```
dtacheny, line 69: Yes, this works! However, I would use dot notation, like `inventory[i].id` instead of bracket notation whenever possible. Your code is still correct, though!

assignments/callbacks.js
```diff
@@ -2,33 +2,78 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 
+// let multiplyNums = (x, y, cb) => {
+//   const product = x * y;
+//   cb(product);
+// }
+
 function contains(item, list, cb) {
   // contains checks if an item is present inside of the given array/list.
   // Pass true to the callback if it is, otherwise pass false.
+  if (list.indexOf(item)) {
```
dtacheny: Line 55:

`list.indexOf(item)` returns `-1` if `item` isn't found in `list`. Otherwise it returns the first index where `item` is found in `list`. This is a problem, because 0 is a valid index, but 0 is also falsey.

If you want to use `indexOf` to answer this, you would have to alter the condition in the `if` block.

For instance, you could do:
```
if (list.indexOf(item) === -1)
```
Then if `list` does _not_ contain `item`, `indexOf` will return `-1` and the condition will pass. Then `cb(false)` will run.

assignments/callbacks.js
```diff
@@ -2,33 +2,78 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+// }
+
 /* STRETCH PROBLEM */
 
 function removeDuplicates(array, cb) {
   // removeDuplicates removes all duplicate values from the given array.
   // Pass the duplicate free array to the callback function.
   // Do not mutate the original array.
+  let noDuplicatedArray = new Set(array);
+  cb(noDuplicatedArray);
```
dtacheny: This works. See if you can figure out how to do it using a `for` loop!

You would also need to convert `noDuplicatedArray` into an array before passing it into the callback, since Sets are not quite the same as Arrays. You could do this by using `Array.from()`: 
```
Array.from(noDuplicatedArray)
```

dtacheny (commented):
Great job 👍

ghost:
@danitacheny 

I have updated my code and also the `for` loop way for `removeDuplicates`. 

( writing tests really helps in preventing careless code :| )

PR #170, opened 2018-06-25
Kamal Mukkamala

Kamalnrf:


natdf:
Excellent job Kamal!
I've noticed that there is some inconsistency in the use of semicolons throughout your code. It can be difficult to keep everything consistent and something like this is easy to overlook. One suggestion would be to use a code formatter such as the one I've mentioned before called Prettier. I think it is also available for Atom. 
Otherwise, great work overall!

PR #174, opened 2018-06-25
Erica Thompson: Open PR

EricaThompson:


assignments/arrays.js
```diff
@@ -2,7 +2,12 @@
 
 // The car dealer has all of their inventory housed in the array seen below.  Scroll down past the data to find out how you can help the car dealer.
 
-let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year":2009},
+let inventory = [{
```
EricHech, line 5: Did you break this first one out to make it more readable? I like it.

assignments/arrays.js
```diff
@@ -63,34 +68,61 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+console.log(`Car 33 is a`, inventory[33]['car_year'], inventory[33]['car_make'], inventory[33]['car_model'] );
```
EricHech, line 71: Try not to use backticks unless you're using template literals. Either you could have passed it all in is one argument with template literals, or you could have just passed in multiple arguments to the console as you did (but then just use normal quotes for that first argument).

assignments/arrays.js
```diff
@@ -63,34 +68,61 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
-let lastCar = 0;
-console.log();
+let lastCar = inventory.length - 1;
+console.log(inventory[lastCar]['car_make'], inventory[lastCar]['car_model']);
```
EricHech, line 78: Very nice. This would be more typical notation, and it's a little more readable:
`inventory[lastCar].car_make`

Same goes for other instances of the same throughout the file.

assignments/arrays.js
```diff
@@ -63,34 +68,61 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
 let carModels = [];
-console.log();
+
+for (let model in inventory){
```
EricHech, line 84: I found that documentation I was looking for here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

Scroll down to: **Difference between for...of and for...in** and try to consume that:

Both for...in and for...of statements iterate over something. The main difference between them is in what they iterate over.

The for...in statement iterates over the enumerable properties of an object, in an arbitrary order.

The for...of statement iterates over data that iterable object defines to be iterated over.

The following example shows the difference between a for...of loop and a for...in loop when used with an Array.

-----------

Object.prototype.objCustom = function() {}; 
Array.prototype.arrCustom = function() {};

let iterable = [3, 5, 7];
iterable.foo = 'hello';

for (let i in iterable) {
  console.log(i); // logs 0, 1, 2, "foo", "arrCustom", "objCustom"
}

for (let i in iterable) {
  if (iterable.hasOwnProperty(i)) {
    console.log(i); // logs 0, 1, 2, "foo"
  }
}

for (let i of iterable) {
  console.log(i); // logs 3, 5, 7
}

assignments/arrays.js
```diff
@@ -63,34 +68,61 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+
+
 
 // ==== Challenge 6 ====
 // A buyer is interested in seeing only BMW and Audi cars within the inventory.  Return an array that only contains BMW and Audi cars.  Once you have populated the BMWAndAudi array, use JSON.stringify() to show the results of the array in the console.
 let BMWAndAudi =[];
-console.log();
+
+for (let each in inventory){
+    if ((inventory[each]['car_make'] === "BMW") || (inventory[each]['car_make'] === "Audi")){
```
EricHech, line 119: You don't need these extra parentheses around each `||` condition.

assignments/callbacks.js
```diff
@@ -2,33 +2,62 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 function multiplyNums(x, y, cb) {
   // multiplyNums multiplies two numbers and passes the result to the callback.
+  cb(x * y);
+
 }
 
 function contains(item, list, cb) {
   // contains checks if an item is present inside of the given array/list.
   // Pass true to the callback if it is, otherwise pass false.
+  if (list.includes(item) == true){
```
EricHech, line 37: `if (list.includes(item) == true) cb(true);`
`else cb(false);`

You could have done it this way, but just to be clear, some companies will require that it is written the way you did because it can prevent errors from happening if people decide to start editing the code by adding in more lines or something. So your way may be considered better.

assignments/callbacks.js
```diff
@@ -2,33 +2,62 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+  } else {
+    cb(false);
+  }
+
 }
 
 /* STRETCH PROBLEM */
 
-function removeDuplicates(array, cb) {
+function removeDuplicates(array) {
```
EricHech, line 47: Woooow, this is a neat way to solve it. I wouldn't have thought of this. Nice job.

assignments/objects.js
```diff
@@ -19,46 +19,140 @@ let example = {
...
+  email: "mmelloy0@psu.edu",
+  gender: "F"
+}
+let kennan = {
+  id: 2,
+  name: "Kennan",
+  email: "kdiben1@tinypic.com",
+  gender: "M",
+  speak: function speak(){
+    phrase = "Hello my name is" + " " + this.name;
```
EricHech, line 34: Good opportunity for those template literals:
`return 'Hello my name is ${this.name}';`

So I used single quotes up there, but they should be backticks. I can't use backticks in here without github formatting them and thinking I'm trying to do something else, lol. Kind of annoying....

assignments/objects.js
```diff
@@ -19,46 +19,140 @@ let example = {
...
 
 // 1. Create a parent object with properties for name and age.  Make the name Susan and the age 70.
 // 2. Nest a child object in the parent object with name and age as well.  The name will be George and the age will be 50.
 // 3. Nest a grandchild object in the child object with properties for name and age.  The name will be Sam and the age will be 30
 // 4. Give each of the objects the ability to speak their names using the this keyword.
 
-let parent = {}
+let child = {};
+let grandchild = {};
+
```
EricHech, line 105: So here you could have done something like:

function speakFunc() {
    return 'I'm ${this.name}';
}

And then in each object, like:

let parent = {
    name: "Susan",
    age: 70,
    speak: speakFunc,
}

Notice that I didn't do `speak: speakFunc()` because that would have called the function and stored the result of `speakFunc()` into `speak` which has a totally different effect. This is what Trevor and I were talking about last night during standup for some time.

EricHech (commented):
Really great job :+1:

PR #175, opened 2018-06-25
Aurel Ngo-Javascript-I

SurrealityDD:


TERR1E (commented):
You were doing great yesterday. Remember to review the link I shared with you.
Here are more helpful links to get you through:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_Objects

https://stackoverflow.com/questions/1184123/is-it-possible-to-add-dynamically-named-properties-to-javascript-object

http://www.dyn-web.com/tutorials/object-literal/properties.php

PR #181, opened 2018-08-06
done object assignment and stretch (doing pull request early

decagondev:
working on the arrays part now


decagondev:
done the array challenges

PR #192, opened 2018-08-06
JavaScript I - Evan Carlstrom

ecarlstrom:


crypto-jones:
I love how you made multiple commits and explicitly described what you did for each one. Great job! 
Your code looks clean & works as expected. One suggestion is try using template literals in those `console.log` statements in `arrays.js` vs concatenation.  Also, in `objects.js` , if you have some free time, try to finish the stretch goal. I think it's really helpful in understanding how methods work as well as how useful they are.

PR #207, opened 2018-08-06
Linda Yang - JavaScript I Stretch

lyang9:


crypto-jones:
Overall, your code is clean, easy to read, and it looks correct. Great job! One thing I would suggest is try using template literals in your console.logs.  Also, I noticed that you wrote your methods outside of the objects themselves. This is ok, but know that you can also put them in the object.

PR #211, opened 2018-08-06
Colin-Preston

ghost:
I had git bash issues but they are now resolved. I also had vs code issues but am currently resolving them now.


crypto-jones:
Overall, your code is clean and readable. Good job! 
I do have two critiques:

1. Be consistent with the spacing of nested blocks of code in the for loop in the last challenge in arrays.js.

2. In objects.js, for the `parent` object, you should have used a method called `speak`. A method is a function inside of an object. Please check the solution code to see what I mean here.

PR #216, opened 2018-08-06
Tristen Neu -- JavaScript-I

ghost:


crypto-jones:
Overall, your code is clean, easy to ready, and works as expected. A couple of tips:

1 . In `arrays.js` for Challenge 1, try using template literals in your return statement, so you don't have to concatenate so much. 

2. In `arrays.js` for Challenge 2, you came up with a very creative solution! However, see how you could simplify your solution into less lines of code & make it more readable.

PR #225, opened 2018-08-06
JS 1 - Joseph Bradley - Team Kepf

jmbradley:
+ Waxed'em.......but definitely took a while to get into the groove.

+ @khepf - Jeremy Kepf


khepf:
Very nice work here. No question, you are ready for Day 2. Great work!

PR #243, opened 2018-08-14
Joseph Thompson - Javascript-I

jthomps838:
@kiadorii

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# JavaScript - I
+# JavaScript - I - Joseph Thompson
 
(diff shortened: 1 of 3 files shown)


kiachoi (commented):
✔️ What Went Well:
---_
+ Properly working array.js  & objects.js
+ Clean commit messages and code
+ Stretch Goals

🎗 Recommendation(s):
----
+ I noticed that you used `var` in your for loops. I would recommend slowly transferring over to let. This is because `let` allows you to declare variables that are limited in scope to the block, statement, or expression on which it is used. While `var` defines it globally regardless of its scope.

Other:
----
+ 3/3

PR #251, opened 2018-08-14
Stefan Clem

iepoch:
@47analogy


47analogy:
All challenges in arrays.js and objects.js are completed successfully and follow consistent style and conventions. Nice addition of adding a variable to keep the car count for Challenge #3 in arrays.js. Great job!

PR #270, opened 2018-08-15
Erin-Ronae Scriber

erscriber:


assignments/objects.js
```diff
@@ -18,27 +18,78 @@ const example = {
 }
 
 // Write your intern objects here:
+const Mitzi  = {
```
kiachoi, line 21: This was intended to be in one array - I will mention in the staff channel to reword it.
Now that you do know haha, how would you switch this?

kiachoi (commented):
✔️ What Went Well:
----
+ Great work on completing Array.js & objects.js!

🎗 Recommendation(s):
----
+ None, it looks good!

PR #275, opened 2018-08-15
Jennifer King - JavaScript I

pghrogue:
### Objects
- [x] Complete each challenge presented before moving on to Arrays.

### Arrays
- [x] The [arrays.js](assignments/arrays.js) assignment takes us through a large data set of used cars.  You have been asked to help a used car business with some customer requests based on their inventory.  Use for loops and arrays to solve their problems.

- [x]  Utilize the the array `inventory` to complete your challenges
- [x]  You are not permitted to use map, reduce, or filter to solve these problems.  Only use a basic for loop.
- [x]  Complete each challenge presented before moving on to stretch.

### Stretch
- [ ]  Move on to tomorrow's content and start studying callbacks, write a few of your own to get the hang of it.
- [ ]  Look at array methods like .map(), .reduce(), .filter(). use them on the data in the arrays assignment to accomplish the same things you did with the ES5 for loop.

### DIFF
--- a/assignments/objects.js
+++ b/assignments/objects.js
@@ -1,7 +1,8 @@
 // Let's get some practice writing a few objects for a new group of interns at a small business.
 
 // ==== Challenge 1: Writing Objects ==== 
-// HR needs some information on the new interns put into a database.  Given an id, email, first name, and gender. Create an object for each person in the company list:
+// HR needs some information on the new interns put into a database.  Given an id, email, first name, and gender. Create an object for 
+// each person in the company list:
 
 // 1,mmelloy0@psu.edu,Mitzi,F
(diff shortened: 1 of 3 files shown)


assignments/objects.js
```diff
@@ -18,27 +19,61 @@ const example = {
 }
 
 // Write your intern objects here:
-
+const mitzi = {
```
kiachoi, line 22: You have this written as separate objects, could you think of another way where these objects are nested into one??

kiachoi (commented):
✔️ What Went Well:
----
+ Great work on arrays.js and objects.js
+ Clean commit messages and code

🎗 Recommendation(s):
----
+ I noted one about nesting the objects into one.

PR #290, opened 2018-08-15
Bruce Cabanayan - Javascript I

bcabanayan:
This is my submission for the Javascript I project.


assignments/objects.js
```diff
@@ -19,27 +19,79 @@ const example = {
 
 // Write your intern objects here:
 
+const mitzi = {
```
kiachoi, line 22: Do you think that you can write these objects in a way that they are nested into one?

assignments/objects.js
```diff
@@ -49,16 +101,50 @@ const example = {
...
+      return this.name;
+    }
+
+    const grandchild = {
+      "name": "Sam",
+      "age": 30,
+      speak: function() {
+        return this.name;
+      } 
+    }
```
kiachoi, line 122: There are extra '{}' that are not matching - please take another look at this when you have time! 😄

kiachoi (approved):
✔️ What Went Well:
----
+ Array.js and object.js functioning properly

🎗 Recommendation(s):
----
+ When I ran `node ...` - I received an error starting at line 111 on objects.js. I made a note of it but once I fixed it on my local machine, the console.logs worked fine.

PR #312, opened 2018-08-15
Jerrard-Smith JavaScript-I WIP

JerrardSmithOfficial:
@47analogy


47analogy:
All challenges in arrays.js and objects.js are completed successfully. For Challenge#1 in arrays.js, remember that arrays are 0 indexed, so car with id of 33 is let selector = inventory[32]. Also, even though we didn't cover this topic until after this assignment, try use const until you can't, then use let for variables (Kennan object in objects.js). Your code is consistent and very readable. Great work on the project and appreciate you stepping up and helping out your fellow teammates. Keep up the good work!

PR #318, opened 2018-08-16
Demetri Bethel JavaScript-I

kingmetri:
@47Analogy


47analogy:
All challenges in arrays.js and objects.js are completed successfully. Nice understanding that arrays are 0 index based for using the correct index to get the last car in the array for Challenge #2 in arrays.js. You can extend this to also to use the inventory length property to get the same results (for example: let lastCar = inventory[inventory.length - 1].car_make;) to get the last item in an array. Nice job of adding an additional variable to keep track of the car count in Challenge #3 of arrays.js. Great job!

PR #320, opened 2018-08-16
javascript-I

ShardeDeLane:


assignments/arrays.js
```diff
@@ -63,34 +63,74 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
-
-
+// console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+for(i = 0; i < inventory.length; i++) {
```
Raymondgrc: I like your work, Sharde. For this problem, we don't need a for loop because we only want value. A for loop makes the computer do more work than it needs to.

assignments/arrays.js
```diff
@@ -63,34 +63,74 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+	if(inventory[i] === inventory[32])
+	console.log(inventory[32].car_make,inventory[32].car_model, inventory[32].car_year);
+}
 
+console.log("*****************");
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
 let lastCar = 0;
-console.log();
+for(i = 0; i <= inventory.length; i++) {
```
Raymondgrc: Same thing here, we could use something like `console.log(arrayName[ArrayName.length - 1];`

assignments/arrays.js
```diff
@@ -63,34 +63,74 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+	}
+}
+
 
 // ==== Challenge 6 ====
 // A buyer is interested in seeing only BMW and Audi cars within the inventory.  Return an array that only contains BMW and Audi cars.  Once you have populated the BMWAndAudi array, use JSON.stringify() to show the results of the array in the console.
 let BMWAndAudi =[];
-console.log();
-
+for (var key in inventory) {
```
Raymondgrc: Nice job using a "Key" "in" loop :)

PR #322, opened 2018-08-17
Brandon G. Elzy JavaScript-I

belzy:
@fireinjun

### DIFF
--- a/assignments/arrays.js
+++ b/assignments/arrays.js
@@ -63,34 +63,47 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+for (let i = 0; i < inventory.length; i++) {
+  if (inventory[i].id === 33) {
+    console.log(`Car 33 is a ${inventory[i].car_year} ${inventory[i].car_make} ${inventory[i].car_model}` );
+  }
+}
 
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
-let lastCar = 0;
-console.log();
+let lastCar = inventory[inventory.length - 1];
+console.log(`${lastCar.car_make} ${lastCar.car_model}`);
 
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
 let carModels = [];
-console.log();
+for (let i = 0; i < inventory.length; i++) {
+  carModels.push(inventory[i].car_model);
+}
+console.log(carModels.sort());
 
 // ==== Challenge 4 ====
 // The accounting team needs all the years from every car on the lot. Create a new array from the dealer data containing only the car years and log the result in the console.
 let carYears = [];
-console.log();
+for (let i = 0; i < inventory.length; i++) {
+  carYears.push(inventory[i].car_year);
+}
+console.log(carYears);
 
 // ==== Challenge 5 ====
 // The car lot manager needs to find out how many cars are older than the year 2000. Using the carYears array you just created, find out how many cars were made before the year 2000 by populating the array oldCars and logging it's length.
 let oldCars =[];
-console.log(); 
+for (let i = 0; i < carYears.length; i++) {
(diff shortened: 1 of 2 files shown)


assignments/arrays.js
```diff
@@ -63,34 +63,47 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+for (let i = 0; i < inventory.length; i++) {
```
designerexpert, line 66: I like that you made this dynamic, most people would have chosen the easy way out selecting index number 32. Good job 👍

assignments/arrays.js
```diff
@@ -63,34 +63,47 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+}
 
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
-let lastCar = 0;
-console.log();
+let lastCar = inventory[inventory.length - 1];
+console.log(`${lastCar.car_make} ${lastCar.car_model}`);
```
designerexpert, line 77: Perhaps leave yourself a little message in the console log like so:
```The last car in the lot is a: ${lastCar.car_make} ${lastCar.car_model}```

File: assignments/arrays.js [hunk 3]
```diff
@@ -63,34 +63,47 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
 let carModels = [];
-console.log();
+for (let i = 0; i < inventory.length; i++) {
+  carModels.push(inventory[i].car_model);
+}
+console.log(carModels.sort());
```
designerexpert, line 85: I like that you used sort here. Have you considered every making your own sorting function? How would you go about doing this? Hint: using charCodeAt() this returns the numerical representation of a string character, and then you can use that to order by numbers since math has real order. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt

assignments/arrays.js
```diff
@@ -63,34 +63,47 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 // ==== Challenge 6 ====
 // A buyer is interested in seeing only BMW and Audi cars within the inventory.  Return an array that only contains BMW and Audi cars.  Once you have populated the BMWAndAudi array, use JSON.stringify() to show the results of the array in the console.
 let BMWAndAudi =[];
-console.log();
-
-
-
+for (let i = 0; i < inventory.length; i++) {
+  (inventory[i].car_make === "Audi" || inventory[i].car_make === "BMW") ? BMWAndAudi.push(inventory[i]) : false;
```
designerexpert, line 107: I like that you used the "or || " inside your conditional ternary statement, but you do not need the parenthesis, however it makes it easier to read! Alternatively you can do "null" after the ternary like so ` :  null`

assignments/objects.js
```diff
@@ -18,47 +18,109 @@ const example = {
...
+    "name": "Mitzi",
+    "email": "mmelloy0@psu.edu",
+    "gender": "F"
+ },
+ {
+    "id": 2,
+    "name": "Kennan",
+    "email": "kdiben1@tinypic.com",
+    "gender": "M",
+    Speak: () => {
```
designerexpert, line 33: I love arrow functions. Have you considered the difference between an arrow function and a regular function? Think about this for the future: https://medium.com/@thejasonfile/es5-functions-vs-es6-fat-arrow-functions-864033baa1a

designerexpert:
## Comments:
* I am glad you finished this assignment, and from the looks of it you grasp the concept very well.
* Perhaps spend some time reviewing the notes I left you to get more insight about the topics.
## Objects:
- [x] Read the instructions found within the file carefully to finish the challenges.
- [x] Complete each challenge presented before moving on to Arrays.
## Arrays:
- [x] Utilize the the array inventory to complete your challenges
- [x] You are not permitted to use map, reduce, or filter to solve these problems. Only use a basic for loop.
- [x] Complete each challenge presented before moving on to stretch.
## Stretch:
- [x] Move on to tomorrow's content and start studying callbacks, write a few of your own to get the hang of it.
- [x] Look at array methods like .map(), .reduce(), .filter(). use them on the data in the arrays assignment to accomplish the same things you did with the ES5 for loop.

PR #325, opened 2018-08-17
Chris Juelfs JavaScript-I completed!

CheapCharlie:


47analogy:
Great job on the project. One thing to note for Challenge #1 in arrays.js is that since arrays are 0 indexed based, the index for car with id 33 is actually inventory[32].  For challenge #3, one way of getting the car makes in the correct order is using a for loop to create a car models array and then using the sort method.  In Challenges #3 in objects.js., since you have an array of objects, you'd want to add the method to the interns array 

```interns.speak = function () {
  return "Hello, my name is" + " " + interns[1]["name"];
}```

47analogy:
Great job on the project. One thing to note for Challenge #1 in arrays.js is that since arrays are 0 indexed based, the index for car with id 33 is actually inventory[32]. For challenge #3, one way of getting the car makes in the correct order is using a for loop to create a car models array and then using the sort method. In Challenges #3 in objects.js., since you have an array of objects, you'd want to add the method to the interns array

```interns.speak = function () {
  return "Hello, my name is" + " " + interns[1]["name"];
}```

47analogy:
Great job on the project. Couple of things to note are that since arrays are 0 indexed, for Challenge#1 in arrays.js, car with id =33 is inventory[32]. For challenge#2, you could use a for loop and push the car models to a new array, which you could apply the array sort method. For Challenge#3 in objects.js, since you created an array of objects, you could add the speak method to the the interns array, i.e. interns.speak = function () {...

CheapCharlie:
Regarding you challenge#2 notes. wouldn't the way i did it be the simpler and/or more efficient way to do it though?

PR #330, opened 2018-08-19
JavaScript -1 Henry Moland

henrymoland:


47analogy:
Nice job! All challenges in arrays.js and objects.js are completed. Just wanted to point out a couple of items:

For Challenge #1, remember that arrays are 0 indexed so that the car with id = 33 is actually inventory[32]

Notice that you use bracket notation a lot in objects.js. While this works, it's better to use dot notation. Bracket notation is useful when the key is a variable -- we know it exists,  but not sure exactly what value is). 

Also in the same file, another way to associate multiplyNums with Antonietta is to add it as a method on the object.

PR #332, opened 2018-09-10
Ryan Boris, Project - JavaScript I

secantsquared:
@jeffyjkang


jeffyjkang:
hey i think the stretch conversion file is not here, not sure what happened but if you pull the file too soon, sometimes the last minute changes that josh does, does not get updated. Also you may want to check the keven email in the object.js file, i think you pulled the wrong data in, but other than that, looks great 3/3! considering, the stretch you did is alittle weird/ wrong file??? not sure what happened exactly o.O

secantsquared:
Fixed the IDs for the objects and fixed the Keven data.
Renamed the two stretch goal files in order to identify them easier.
Note: The third stretch goal is completed and included at the end of the objects.js assignment file.

I think that fixes up the raised issues - thank you for catching those!
@jeffyjkang

PR #334, opened 2018-09-10
Adam Hinckley

adamhinckley:


assignments/arrays.js
```diff
@@ -63,34 +314,52 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
-
+console.log(`Car 33 is a: ${inventory[32].car_year}, ${inventory[32].car_make}, ${inventory[32].car_model}`);
```
Velsu, line 317: This is good solution, but safer method would be to loop thru the array and compare id's. Lists can change in the future and we can't be sure that car on idex 33 will have same id.

assignments/objects.js
```diff
@@ -49,7 +94,26 @@ const example = {
 // 3. Nest a grandchild object in the child object with properties for name and age.  The name will be Sam and the age will be 30
 // 4. Give each of the objects the ability to speak their names using the this keyword.
 
-const parent = {}
+const parent = {
+  name: 'Susan',
+  age: 70
```
Velsu, line 99: Missed a coma here, but even so there is an issue with nesting here. What i would do is create a spearate function for speak and just add reference to it in an object like `speak: speak` and name child and grandchild objects.

Velsu (commented):
Great job Adam. Everything looks good. There were some small issues and bugs but i marked and commented them below.

PR #339, opened 2018-09-10
Samuel Machat

axolotl:


assignments/objects.js
```diff
@@ -11,54 +11,176 @@
...
 }
 
 // Write your intern objects here:
+;`1,mmelloy0@psu.edu,Mitzi,F,
+2,kdiben1@tinypic.com,Kennan,M,
+3,kmummery2@wikimedia.org,Keven,M
+4,gmartinson3@illinois.edu,Gannie,M
+5,adaine5@samsung.com,Antonietta,F`
+  .split('\n')
+  .map(chunk => chunk.split(','))
```
mmargo10, line 27: Nice use of .map, but the the assignment called for using for loops and not using .map, .filter,  .reduce. These were outside of the scope of the assignment.

mmargo10 (commented):
Your code is very clear and concise, however you did step outside of the requirements for the assignment a bit.

PR #344, opened 2018-09-10
Edwin Torres

ETorresM2:


assignments/arrays.js
```diff
@@ -72,25 +83,59 @@ console.log(`Car 33 is a *car year goes here* *car make goes here* *car model go
 let lastCar = 0;
 console.log();
 
+let cMake = "";
```
Velsu, line 86: You are declaring same variables again which will cause conflict.

assignments/arrays.js
```diff
@@ -72,25 +83,59 @@ console.log(`Car 33 is a *car year goes here* *car make goes here* *car model go
...
+let cModel = "";
+let cYear = "";
+for (i=0; i<inventory.length; i++) {
+  if (inventory[i].id === inventory[inventory.length-1].id)
```
Velsu, line 90: No need to do the loop just to get last item in the array, faster solution would be to grab it by last index. We You do not know how many items are in the array just use array.length -1 to get the last item.

So for example:
`let lastCar = inventory[inventory.length -1]`

And then You can do `lastCar.car_model` etc.

Velsu (commented):
Great job on the assingment Edwin. You have properly completed arrays and objects exercises. There are some issues with variable declarations that i mentioned in the comments below, along with some unecessary moves, but overall You did good job.

Your git workflow has improved a little bit, but it could still be more descriptive.

PR #346, opened 2018-09-10
JavaScript-I - Blake Koennecker

koennecker:


assignments/arrays.js
```diff
@@ -63,34 +63,53 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
-
+console.log(`Car ${inventory[32].id} is a ${inventory[32].car_year} ${inventory[32].car_make} ${inventory[32].car_model}`);
```
Velsu, line 66: Other, more safer aproach would be to loop thru the whole collection and look for that id. Just in case data would shift/change in the future we will still have a safe way to determine which car is with id of 33.

Velsu (commented):
Excellent job Blake, all the challenges solved successfully. Also great git workflow.

PR #348, opened 2018-09-10
Jake Thomas

jmaxt12:


mmargo10 (commented):
I would suggest taking advantage of the resources in the training kit to strengthen your understanding of objects and arrays. Keep up the good work!

PR #350, opened 2018-09-10
Carlo Clamucha - JavaScript I

CarloC24:
Finished MVP and Stretch


KevinBrack:
arrays.js line 100 has a syntax error `for(let i = 0; i < inventory.length, i++){` has a comma instead of semicolon.

objects.js line 97 needs a comma separating the two objects in the array. Also the method is catching undefined somewhere.

Other than that looking good

PR #361, opened 2018-09-10
objects

codypeak:


KevinBrack:
arrays.js challenge 4 has a comma instead of a semicolon in the for loop. Challenge 5 is missing the for loop but using [i] for the index anyway. Other than that it is looking good sir

PR #363, opened 2018-09-10
JavaScript-I By - Michael Checo

MichaelCheco:


KevinBrack:
All console logs are as expected. Code compiled and ran without errors. Nice work!

PR #367, opened 2018-09-10
JavaScript-I Brian Gregorius

brianmgre:
@CookieMonsta89


CookieMonsta89:
Nice job.  Your code works as required.  Only recommendation is to put your console.log's beneath the function.  When you call it before, it hasn't had the chance to run the code.  I had to move them below the function to get the function to work.

PR #375, opened 2018-09-10
Javascript

Bryan-GF:
@Jeffyjkang


jeffyjkang:
hey bryan great stuff, you may want to recheck the multiplynums method, i get an undefined before console.logging the 12

PR #381, opened 2018-09-10
edward jeong

edjeong99:


assignments/objects.js
```diff
@@ -17,29 +17,76 @@ const example = {
...
+// Write your intern objects here:
+const companyList = [
+  {
+    "id": 1,
+    "name": "Mitzi",
+    "email": "mmelloy0@psu.edu",
+    "gender": "F"
+
+  },
+  KennanObj,
```
Velsu, line 31: JS will throw that KennanObj is undefined. I would not rely on hoisting so much in JS as it is mainly functions that get  hoisted and not variables. Quick fix is moving the KennanObj that was defined at the lower level before the companyList which will fix the issue.

Velsu (commented):
Great job on the assingments Edward. You successfully completed all challenges. You are displaying  understanding of current subject.

Git workflow is fluid and detailed, nice job 👍

PR #384, opened 2018-09-10
Oleksandr Stasyev

CSDisciple:
Partially finished arrays.js, #3 isnt done yet!


CookieMonsta89:
Good job Olek! Most of your code works as required.  Regarding the carmodels challenge, you could use either or these carModels.push(inventory[i]['car_model'])
  cardModels.push(inventory[i].car_model).

PR #387, opened 2018-09-10
Shawn Antonucci - JavaScript-I

shawnantonucci:
@https://github.com/kamry-bowman

### DIFF
--- a/assignments/arrays.js
+++ b/assignments/arrays.js
@@ -79,7 +79,7 @@ for (i = 0; i < inventory.length; i++) {
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
 let lastCar = inventory[inventory.length-1];
-//console.log(lastCar.car_make, lastCar.car_model);
+console.log(lastCar.car_make, lastCar.car_model);
 
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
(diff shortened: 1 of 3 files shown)


assignments/arrays.js
```diff
@@ -63,34 +63,62 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
 
+//for loop, if inventory[i] = 32
+
+for (i = 0; i < inventory.length; i++) {
```
kamry-bowman, line 69: use `for (let i = 0` not `for (i = 0`. The second one is an implicit assignment to global scope, which is not good.

assignments/arrays.js
```diff
@@ -63,34 +63,62 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+    carModels.push(inventory[i].car_model);
+    carModels.sort();
+}
+console.log(carModels);
 
 // ==== Challenge 4 ====
 // The accounting team needs all the years from every car on the lot. Create a new array from the dealer data containing only the car years and log the result in the console.
 let carYears = [];
-console.log();
+for(let i = 0; i < inventory.length; i++) {
```
kamry-bowman, line 96: I'd recommend to put a space between for and (. I would say only have parentheses right up against a word if you're doing function invocations

kamry-bowman (commented):
Looks great! Only minor comments from me!

PR #388, opened 2018-09-10
Phil Hart

PhilHart23:


assignments/objects.js
```diff
@@ -18,28 +18,71 @@ const example = {
...
+  "gender":"M"
+}
+
+const Antonietta = {
+  "id":"5",
+  "name":"Antonietta",
+  "email":"adaine5@samsung.com",
+  "gender":"F"
+}
+
```
mmargo10, line 55: You probably want to use let instead of const for all of your intern objects. Const is immutable, meaning that it is not changeable and you cannot reassign to the const variable. With a let variable, it is mutable and will allow for reassignment.

mmargo10 (commented):
I would suggest firming up your understanding of objects and arrays with the resources in the training kit. Keep up the good work!

PR #399, opened 2018-09-10
Katie Fitzpatrick - JavaScript-I

KatieFizzy:


assignments/arrays.js
```diff
@@ -57,40 +57,56 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 // ==== Challenge 5 ====
 // The car lot manager needs to find out how many cars are older than the year 2000. Using the carYears array you just created, find out how many cars were made before the year 2000 by populating the array oldCars and logging it's length.
-let oldCars =[];
-console.log(); 
-
+let oldCars =[]; 
+for (let i = 0; i < carYears.length; i++) {
+    if (carYears[i] < 2000)
+    oldCars.push(carYears[i]);
```
kamry-bowman, line 98: I would strongly recommend always using blocks for if statements.  It prevents you forgetting and adding additional lines afterwards, but because the multi-lines aren't in a block, it won't behave correctly.

The pattern I would recommend is only do the following:
`if (condition) behavor // one line`
```
if (condtion) {
 behavior
} // multiline
```
Some people think you should even include curly brackets on the single line one. (I tend to do that as well). Just my two cents..

assignments/objects.js
```diff
@@ -49,16 +89,38 @@ const example = {
...
 
 // Log the parent object's name
+console.log(parent["name"]);
 
 // Log the child's age
-
+console.log(parent.child.age);
 // Log the name and age of the grandchild
-
+console.log((parent.child.grandchild.name)+''+(parent.child.grandchild.age));
```
kamry-bowman, line 120: you could also just do `console.log(parent.child.grandchild.name, parent.child.grandchild.age);`

kamry-bowman (commented):
Great job, everything looks great here! Just minor comments

PR #401, opened 2018-09-10
Barbara Hunter

cmdncode:


assignments/arrays.js
```diff
@@ -63,33 +63,54 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+    oldCars.push(carYears[i]);
+  }
+}
+console.log(oldCars.length); 
 
 // ==== Challenge 6 ====
 // A buyer is interested in seeing only BMW and Audi cars within the inventory.  Return an array that only contains BMW and Audi cars.  Once you have populated the BMWAndAudi array, use JSON.stringify() to show the results of the array in the console.
 let BMWAndAudi =[];
+for (let i = 0; i < inventory.length; i++){
+  if (inventory[i].car_make === "Audi" || inventory[i].car_make === "BMW" )
```
Velsu, line 112: It seems the latter part of the code is missing which causes errors. All thats left is to push the found car to new array.

assignments/objects.js
```diff
@@ -10,35 +10,84 @@
...
+console.log (Antonietta.gender);
+
 // ==== Challenge 3: Object Methods ==== 
 // Give Kennan the ability to say "Hello, my name is Kennan!" Use the console.log provided as a hint.
-// console.log(kennan.speak());
+Kennan.speak = function (){
+  console.log ("Hello my name is Kennan!!!!");
+}
+
+console.log(kennan.speak());
```
Velsu, line 83: kennan here should be the uppercase - `Kennan` or it will throw an error.

assignments/arrays.js
```diff
@@ -63,33 +63,54 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
-
+// 
+for (let i = 0; i < inventory.length; i++) {
+  if (inventory[i].id === 33) {
```
Velsu, line 68: Great solution. Plenty of people will go the easy way and just use index from the array, but if that array would be subject to change in the future, Your code will still work.

Velsu (commented):
Barbara great job on the assignments. You missed last exercise on arrays but i believe it was by mistake, because Your code hints that You knew what You are doing.

As for objects, there is a typo for kennan, but except that everything looks solid, good job 👍

PR #404, opened 2018-09-10
Nick Kruger - Javascript I

NickKruger:


assignments/arrays.js
```diff
@@ -63,7 +63,12 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+
+for(let i = 0; i === inventory["id" : "33"; i++){
```
Ta1grr: In Challenge 1, it wants you to replace the *car year goes here*, *car make goes here*, and *car model goes here*. In layman term, you want to call the variable, and select the object with an id of 33, and output the name it wants in each spots.

assignments/arrays.js
```diff
@@ -63,7 +63,12 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+
+}
+
+console.log(`Car 33 is a (inventory*car year goes here* *car make goes here* *car model goes here*` );
```
Ta1grr: Challenge 2, It wants you to select the last car of the array, You can hard code it by writing inventory[49], but given the if in the future they add more cars into the inventory, then it won't be the last one anymore. So we know by counting the object inside the array, it is 50. But we know computers count by based 0, so think of the "difference" between the two and put that logic into inventory[] so it can select the last one every time.

assignments/objects.js
```diff
@@ -19,27 +19,80 @@ const example = {
...
 // Gannie's name
 
+console.log(employeeFour["name"]);
+
 // Antonietta's Gender
 
+console.log(employeeFive["gender"]);
+
+
 // ==== Challenge 3: Object Methods ====
```
Ta1grr, line 84: In this Challenge, it wants you to add a method into the kennan(employeeTwo) you can assigned a name into an object by doing employeeTwo.speak then assign a function to it. 
employeeTwo.speak = function() {
 (Your block of code here)
}

assignments/objects.js
```diff
@@ -19,27 +19,80 @@ const example = {
 
 // Write your intern objects here:
 
+const employeeOne = {
```
Ta1grr, line 22: Great job on creating objects, and the naming convention really help distinguish between each object.

assignments/objects.js
```diff
@@ -19,27 +19,80 @@ const example = {
...
+
 // ==== Challenge 3: Object Methods ==== 
 // Give Kennan the ability to say "Hello, my name is Kennan!" Use the console.log provided as a hint.
 // console.log(kennan.speak());
 
+console.log("Hello, my name is " + (employeeTwo["name"])+"!");
+
+
 // Antonietta loves math, give her the ability to multiply two numbers together and return the product. Use the console.log provided as a hint.
 //console.log(antonietta.multiplyNums(3,4));
```
Ta1grr, line 93: So, the second part of this challenge wants you to build upon what you've done above, so same instance, you call: antonietta.multiplyNums = function (two params goes in here) {
  (Your block of code goes in here);
}
(Then console.log the object method and invoke it with two inputs for the two parameters.)

Ta1grr (commented):
## Great
 - Keep chipping away at it, struggling is part of the process, and it forces you to think and learn at the same time.
## Requested Improvements
 - Reach out if you're stuck longer than 20 minutes, CS14 help channel might have answers or people asking the same question. If you can't find what you need I'm also here to help.
## Questions
 - Take each Challenge and break it down into smaller steps.
## Rating: {1-3}
 1 (Does not understand, does not meet expectations)

PR #405, opened 2018-09-10
js1-sgear

sabrinafilipelli:


KevinBrack:
Code compiled and ran without errors. All console logs match expected output. Nice work

PR #406, opened 2018-09-10
JavaScript-I by Katy Hollobaugh

KHollobaugh:


KevinBrack:
On arrays.js challenge 5 line 103 the oldCars array was never defined. Everything else worked as expected. Nice work!

PR #407, opened 2018-09-10
Wanda Evans - Javascript-I

wandadotdev:


jeffyjkang:
hey Wanda good job on the arrays, you may want to go over the object.js file though, you can make them into an array to do the assignment, but it may be easier to just make individual objects by assigning the variable name to the object, (i think this way was the intended way). when i replit your file right now, i dont think you are accessing the information you want.

PR #410, opened 2018-09-10
JavaScript-I By: Yusuf Nafey

yusufnafey:


KevinBrack:
When you are already inside an object you can not use the const keyword. Creating a key of the key-value pair takes the place of this. A value can be a string, number, Boolean, function, or another object.

```{
example1:"string",
 example2: {key:"value"}
}```

```const parent = {
  "name": "Susan",
  "age": 70,
  const child = {
    "name": "George",
    "age": 50,
    const grandchild = {
      "name": "Sam",
      "age": 30
    }
  }
}```

but of course that was a stretch task

Hope this helps. Great work.

PR #412, opened 2018-09-10
Finished Objects

yinwil27:


KevinBrack:
Both files fail to compile due to syntax errors. Can not test results.

PR #423, opened 2018-10-15
James Page: Stretch Complete

jamespagedev:


assignments/arrays.js
```diff
@@ -7,7 +7,7 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 {"id":3,"car_make":"Land Rover","car_model":"Defender Ice Edition","car_year":2010},
 {"id":4,"car_make":"Honda","car_model":"Accord","car_year":1983},
 {"id":5,"car_make":"Mitsubishi","car_model":"Galant","car_year":1990},
-{"id":6,"car_make":"Audi","car_model":"riolet","car_year":1995},
+{"id":6,"car_make":"Audi","car_model":"Riolet","car_year":1995},
```
szincone, line 10: - It's a typo on the assignment (should be Cabriolet), I did the same thing.

assignments/arrays.js
```diff
@@ -63,34 +63,75 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+function getInventoryCar(carId) {
+  for (i = 0; i < inventory.length; i++){
+    if (inventory[i].id === carId){
+      return inventory[i];
+    }
+  }
 
+  return "car ID: " + carId + " not found";
+}
+console.log(`Car 33 is a ` + getInventoryCar(33).car_year + ' ' + getInventoryCar(33).car_make + ' ' + getInventoryCar(33).car_model);
```
szincone, line 75: - Looks like your using back-ticks here, might as well use the template literal syntax 
`Car 33 is a ${getInventoryCar(33).car_year} ${getInventoryCar(33).car_make} ${getInventoryCar(33).car_model} .` (make sure you keep it wrapped in back-ticks.

assignments/arrays.js
```diff
@@ -63,34 +63,75 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+for (let i = 0; i < carYears.length; i++){
+  if (carYears[i] < 2000) {
+    oldCars.push(carYears[i]);
+  }
+}
+console.log(oldCars.length + ' cars are older than 2000');
+
+//// Stretch use reduce()
+let totalNumOldCars = 0;
+carYears.reduce(function (totalOldCars, carYear) {
```
szincone, line 118: I don't know if I'd use reduce here instead of filter. But if you want to use reduce you'd want to do something like this 
- `let count = inventory.reduce(function(n, val) {
    return n + (val.car_year < 2000);
}, 0);`

assignments/stretch-function-conversion.js
```diff
@@ -1,23 +1,36 @@
...
+addOtherFunc = (param1, param2) => param1 + param2; // option 2
+console.log(addOtherFunc(1, 2));
 
 // let subtract = function (param1, param2) {
 //   return param1 - param2;
 // };
 // subtract(1,2);
+let subtract = (param1, param2) => { return param1 - param2}; // option 1
+console.log(subtract(1, 2));
+subtractOtherFunc = (param1, param2) => param1 - param2; // option 2
```
szincone, line 26: - Good job on listing both single line syntax options. Line 26 is way cleaner when you only have a single line.

szincone (commented):
- Very strong stretch, good job implementing filter and map.

PR #425, opened 2018-10-15
David Freitag.

freitagdavid:


assignments/arrays.js
```diff
@@ -62,35 +315,76 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+// The dealer can't recall the information for a car with an id of 33 on his
+// lot. Help the dealer find out which car has an id of 33 by logging the car's
+// year, make, and model in the console log provided to you below:
 
+function getCar(inventory, carId) {
+    for (i = 0; i < inventory.length; i++) {
+        if (inventory[i].id == carId) {
+            let car = inventory[i];
+            return `Car ${carId.toString()} is a ${car.car_year.toString()} ${
```
johnoro, line 326: Being explicit is good, but you actually don't have to convert your numbers to be strings here (especially since using a template literal is already doing that for you):
```javascript
 `Car ${carId} is a ${car.car_year} ${ ...`
```

assignments/arrays.js
```diff
@@ -62,35 +315,76 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+);
 
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
 let carModels = [];
-console.log();
+for (let i = 0; i < inventory.length; i++) {
+    carModels.push(inventory[i].car_model);
+}
+carModels = carModels.sort();
```
johnoro, line 352: You don't need to reassign `carModels` here; the `sort` [method alters the array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) it's called on.

johnoro (commented):
I can tell you used prettier; good on you. Awesome work!

assignments/arrays.js
freitagdavid: Sorry my python showed.

PR #426, opened 2018-10-15
Grace Kang

bananaboots:


assignments/arrays.js
```diff
@@ -63,34 +63,47 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
-
+console.log(`Car 33 is a ${inventory[32]["car_year"]} ${inventory[32]["car_make"]} ${inventory[32]["car_model"]}`);
```
johnoro, line 66: It'd probably be better to just use a variable here to store `inventory[32]` (so that we aren't looking it up multiple times). For example:
```javascript
const car33 = inventory[32];
console.log(`Car 33 is a ${car33.car_year} $car33.car_make} ${car33.car_model}`);
```

assignments/arrays.js
```diff
@@ -63,34 +63,47 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
 let lastCar = 0;
-console.log();
+console.log(`${inventory[lastCar]["car_make"]} ${inventory[lastCar]["car_model"]}`);
```
johnoro, line 72: This isn't the last car (it's the first)!

assignments/arrays.js
```diff
@@ -63,34 +63,47 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
 let carModels = [];
-console.log();
+for (let i = 0; i < inventory.length; i++) {
+    carModels.push(inventory[i]["car_model"]);
+    carModels.sort();
```
johnoro, line 79: Using `.sort` on every iteration probably isn't the best idea for this problem; the JS engine is most likely doing fine with it, but you might want to move it to down below (after the `for` loop).

assignments/stretch-function-conversion.js
```diff
@@ -1,23 +1,29 @@
 // Take the commented ES5 syntax and convert it to ES6 arrow Syntax
 
 // let myFunction = function () {};
+() => {};
 
 // let anotherFunction = function (param) {
 //   return param;
 // };
+(param) => param;
```
johnoro, line 9: single parameters can lose the parentheses (given that you aren't doing anything fancy):
```javascript
const func = param => param;
```

johnoro (commented):
Great work!

PR #427, opened 2018-10-15
Jonathan Heinz

jheinz1999:


assignments/objects.js
```diff
@@ -19,46 +19,161 @@ const example = {
...
+
+const kennan = {
+
+  id: 2,
+  name: "Kennan",
+  email: "kdiben1@tinypic.com",
+  gender: "M",
+  speak: function() {
+
+    return "Hello, my name is " + this.name + "!";
```
johnoro, line 39: You could use [template literal syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) here:
```javascript
```

assignments/stretch-function-conversion.js
```diff
@@ -2,22 +2,39 @@
 
 // let myFunction = function () {};
 
+let myFunction = () => {};
+
 // let anotherFunction = function (param) {
 //   return param;
 // };
 
+let anotherFunction = (param) => param;
```
johnoro, line 11: You're able to drop the parentheses if you have exactly one parameter (if you aren't doing anything fancy with it):
```javascript
let anotherFunction = param => param;
```

assignments/arrays.js
```diff
@@ -63,34 +63,68 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
 
+let car33 = inventory.find((item) => {
+  return item.id === 33
+});
```
johnoro, line 69: I love the use of `find`; shortening this with fancy arrow function conveniences would ultimately give you:
```javascript
let car33 = inventory.find(item => item.id === 33);
```

johnoro (commented):
Great job!

PR #432, opened 2018-10-15
First Submission - all initial challenges completed not including stretch tasks

gittc100:


assignments/objects.js
```diff
@@ -18,28 +18,70 @@ const example = {
...
-// Gannie's name
+// ==== Challenge 2: Reading Object Data ==== 
+// Once your objects are created, log out the following requests from HR into the console:
 
-// Antonietta's Gender
+console.log(mitzi["name"]);
+console.log(kennan["id"]);
+console.log(keven["email"]);
+console.log(gannie["name"]);
+console.log(antonietta['gender']);
```
johnoro, line 63: Keep in mind that bracket notation isn't necessary here; you could've used dot notation like this:
```javascript
console.log(keven.email);
```

assignments/arrays.js
```diff
@@ -63,34 +63,65 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 // ==== Challenge 4 ====
 // The accounting team needs all the years from every car on the lot. Create a new array from the dealer data containing only the car years and log the result in the console.
 let carYears = [];
-console.log();
+for(let i = 0; i < inventory.length; i++){
+    carYears.push(inventory[i]["car_year"]);
+}
+for(let i = 0; i < carYears.length; i++){
+    console.log(carYears[i]);
+}
```
johnoro, line 101: You could log the cars as you push them here (so that you don't have to loop twice):
```javascript
for(let i = 0; i < inventory.length; i++){
    carYears.push(inventory[i]["car_year"]);
    console.log(carYears[i]);
}
```

assignments/arrays.js
```diff
@@ -63,34 +63,65 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+
 
 // ==== Challenge 5 ====
 // The car lot manager needs to find out how many cars are older than the year 2000. Using the carYears array you just created, find out how many cars were made before the year 2000 by populating the array oldCars and logging it's length.
 let oldCars =[];
-console.log(); 
+for(let i = 0; i < carYears.length; i++){
+    if(carYears[i] < 2000){
+        oldCars.push(carYears);
```
johnoro: This isn't pushing the car years! It's instead pushing the whole `carYears` array multiple times, so the length still adds up. Use `carYears[i]`.

assignments/stretch-function-conversion.js
```diff
@@ -1,23 +1,48 @@
...
 //   return num * 3;
 // });
-// console.log(triple);
\ No newline at end of file
+// console.log(triple);
+
+    exampleArray = [1,2,3,4];
+    const triple = exampleArray.map((num) => {
+        return num * 3;
+    });
```
johnoro: Since you only have a single statement/expression that you're returning in your `map` callback, you could shorten it (by loosing the braces and the `return` keyword):
```javascript
    const triple = exampleArray.map((num) => num * 3);
```

You can also lose the parentheses if you only have a single parameter (and you're not doing anything fancy with it):
```javascript
    const triple = exampleArray.map(num => num * 3);
```

johnoro (commented):
Awesome work!

PR #434, opened 2018-10-15
Tommy Huynh

tommaay:
Completed the objects.js challenges including the stretch goals

### DIFF
--- a/assignments/arrays.js
+++ b/assignments/arrays.js
@@ -65,7 +65,7 @@ let inventory = [
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+// console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
 
 console.log(`Car 33 is a ${inventory[32]['car_year']} ${inventory[32]['car_make']} ${inventory[32]['car_model']}`)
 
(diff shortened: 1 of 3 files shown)


assignments/arrays.js
```diff
@@ -63,34 +65,73 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
-
+// console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
 
+console.log(`Car 33 is a ${inventory[32]['car_year']} ${inventory[32]['car_make']} ${inventory[32]['car_model']}`)
```
Kamalnrf, line 70: Never hard type numbers instead you can use the length property.

assignments/arrays.js
```diff
@@ -63,34 +65,73 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+console.log(`${lastCar.car_make} ${lastCar.car_model}`);
 
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
 let carModels = [];
-console.log();
+
+// I did this the hard way first since I thought we couldn't use .sort()... sigh...
+    // const order = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+    // let sortArr = function() {
```
Kamalnrf, line 83: Major problem is that in this sort function you are comparing only one value if there are more than one car model which start with the same letter.

Ideally, this loop is iterating 500 times, which is not bad but there are better ways to sort

check this to read more about it https://medium.com/@rwillt/implementing-sorting-algorithms-in-javascript-b08504cdf4a9

PR #435, opened 2018-10-15
Daily Assingment

holdeelocks:
@Synch3


Synch3:
Great job today Holden!

In the objects.js file I see you added your speak and multiply functions directly to your objects. Well done! Next time it might be worth practicing how to add a key-value pair to an object using dot and bracket notation.

PR #436, opened 2018-10-15
Yasirah Krueng

ykrueng:


assignments/arrays.js
```diff
@@ -62,35 +63,100 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+for (let i=0; i < inventory.length && car33 !== {}; i++) {
+  if (inventory[i].id === 33) {
+    car33 = inventory[i];
+  }
+}
+
+console.log(`Car 33 is a ${car33.car_year} ${car33.car_make} ${car33.car_model}` );
+
+// Stretch covert to .filter
+let car33Filter = inventory.filter(e => e.id === 33)[0];
```
ashwins93, line 79: Good job using the `filter` function to get the object. Try using the `find` function.

assignments/arrays.js
```diff
@@ -62,35 +63,100 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 // ==== Challenge 1 ====
+console.log('\n\n==== Challenge 1 ====');
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+let car33 = {}; // initialize to an empty object
+
+for (let i=0; i < inventory.length && car33 !== {}; i++) {
+  if (inventory[i].id === 33) {
+    car33 = inventory[i];
```
ashwins93, line 72: Once you have found the car you can actually `break` out of the loop to exit early. This saves you a few iterations.

assignments/arrays.js
```diff
@@ -62,35 +63,100 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+console.log(`Stretch using .filter: Car 33 is a ${car33Filter.car_year} ${car33Filter.car_make} ${car33Filter.car_model}`)
 
 
 
 // ==== Challenge 2 ====
+console.log('\n\n==== Challenge 2 ====');
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
-let lastCar = 0;
-console.log();
+let lastCar = inventory[inventory.length-1];
```
ashwins93, line 87: Good use of the variable. The `0` confused a lot of people to set it to the index instead of the object itself.

assignments/arrays.js
```diff
@@ -62,35 +63,100 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 // ==== Challenge 3 ====
+console.log('\n\n==== Challenge 3 ====');
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
 let carModels = [];
-console.log();
+for (let i=0; i < inventory.length; i++) {
+  carModels.push(inventory[i].car_model);
+}
+carModels = carModels.sort();
```
ashwins93, line 97: This is not necessary as `sort` is a mutating method meaning it changes the array in-place. Just `carModels.sort()` would have done the trick. Fortunately sort also returns the original array so this works too.

assignments/arrays.js
```diff
@@ -62,35 +63,100 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+console.log(carModels);
+
+// Stretch covert to .map
+let carModelsMap = inventory.map(e => e.car_model).sort();
```
ashwins93, line 101: Love these one-liners. Concise and powerful.

assignments/arrays.js
```diff
@@ -62,35 +63,100 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
+console.log(oldCars.length); 
+
+// Stretch covert to .filter
+let oldCarsFilter = inventory.filter(e => e.car_year < 2000)
+console.log('Stretch using .filter:');
+console.log(oldCarsFilter.length);
+
+// Stretch covert to .reduce:
+console.log('Stretch using .reduce:');
+console.log(inventory.reduce((s, a) => {
```
ashwins93, line 138: Very good, intuitive use of `reduce`. By using ternary your solution could be made shorter. Check the following code.

```js
inventory.reduce((total, car) => car.car_year < 2000 ? ++total : total)
```

Note the use of pre-increment operator, if you use the post-increment operator this wouldn't work. Can you tell me why is that?

ashwins93 (commented):
As always, amazing job going for all the stretches. Keep working on those code challenges.


**Tip** : Make `git commit`s often with meaningful commit messages.

assignments/arrays.js
ykrueng: I did break early since I check for `car33 !== {}` in the for loop condition, but I think checking it in the body and break does look cleaner.

assignments/arrays.js
ykrueng: ++total, add first then return the value. 
total++, return the current value and add for the next iteration? 💃

PR #438, opened 2018-10-15
Brandon Campbell

BrandCam:


assignments/objects.js
```diff
@@ -18,28 +18,66 @@ const example = {
...
+  "email": "mmelloy0@psu.edu",
+  "gender": "F"
+}
+const Kennan = {
+  "id": 2,
+  "name": "Kennan",
+  "email": "kdiben1@tinypic.com",
+  "gender": "M",
+  "speak": function() {
+    return "Hello, my name is" + " " + this.name + "!";
```
johnoro, line 33: This could be simplified like this:
```javascript
return "Hello, my name is " + this.name + "!";
```

or, by using a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals):
```javascript
```

assignments/objects.js
```diff
@@ -18,28 +18,66 @@ const example = {
...
+console.log(Kennan["id"]);
 // Keven's email
-
+console.log(Keven["email"]);
 // Gannie's name
-
+console.log(Gannie["name"]);
 // Antonietta's Gender
-
+console.log(Antonietta["gender"]);
```
johnoro, line 72: You're able to use dot notation here! For example:
```javascript
console.log(Gannie.name);
```

assignments/stretch-function-conversion.js
```diff
@@ -1,23 +1,28 @@
 // Take the commented ES5 syntax and convert it to ES6 arrow Syntax
 
 // let myFunction = function () {};
-
+let myFunction = () => {};
 // let anotherFunction = function (param) {
 //   return param;
 // };
-
+let anotherFunction = (param) => param;
```
johnoro, line 8: You're able to drop the parentheses if you have exactly one parameter (and aren't doing anything too fancy):
```javascript
 let anotherFunction = param => param; 
```

johnoro (commented):
Great job!

PR #444, opened 2018-10-15
Kevin Marker

kmarker1101:


assignments/arrays.js
```diff
@@ -63,34 +63,50 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+console.log(`Car 33 is a ${inventory[32]["car_year"]} ${inventory[32]["car_make"]} ${inventory[32]["car_model"]}` );
```
johnoro, line 66: It'd probably be better to just use a variable here to store `inventory[32]` (so that we aren't looking it up multiple times). For example:
```javascript
const car33 = inventory[32];
console.log(`Car 33 is a ${car33.car_year} $car33.car_make} ${car33.car_model}`);
```

assignments/arrays.js
```diff
@@ -63,34 +63,50 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
-let lastCar = 0;
-console.log();
+let lastCar = inventory.length - 1;
+console.log(`${inventory[lastCar]['car_make']} ${inventory[lastCar]['car_model']}`);
```
johnoro, line 73: You could shorten this like so:
```javascript
let lastCar = inventory[inventory.length - 1];
console.log(`${lastCar['car_make']} ${lastCar['car_model']}`);
```

johnoro (commented):
Awesome job!

PR #445, opened 2018-10-15
Joseph Eastman

josepheastman:


Synch3:
Nice work Joesph!

In our meeting today we talked about using the sort method as a higher order function, and we went over targeting a specific key value pair in an array of objects.

assignments/objects.js
```diff
@@ -95,20 +95,41 @@ console.log(Antonietta.multiplyNums(3,4));
 // 2. Nest a child object in the parent object with name and age as well.  The name will be George and the age will be 50.
 // 3. Nest a grandchild object in the child object with properties for name and age.  The name will be Sam and the age will be 30
 // 4. Give each of the objects the ability to speak their names using the this keyword.
+const parent = {
+  name: 'Susan',
+  age: 70,
+  speak: function() {console.log(`Hello, my name is ${this.name}!`)},
```
ashwins93, line 101: Object literals are convenient for adding methods such as these. It definitely adds to readability. Makes it easier to see which methods belong to which objects. I would advise to format your function so that it breaks over several lines though, again for the purpose of readability.

assignments/objects.js
```diff
@@ -18,26 +18,72 @@ const example = {
 }
 
 // Write your intern objects here:
+const Mitzi = {
```
ashwins93, line 21: It is a convention to name variables using camel casing. Usually PascalCasing is used only for constructor functions. This code works nonetheless. Following conventions  will make documentations friendly, for instance if you look at the MDN docs you can easily guess that anything with a PascalCasing is a function (constructor function) and those with camelCasing are either methods or properties.

assignments/arrays.js
```diff
@@ -74,16 +74,44 @@ console.log(`${inventory[49]['car_make']} ${inventory[49]['car_model']}`);
 
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
+let carModels = [];
+let modelSort = inventory.sort(function(a,b) {
```
ashwins93, line 78: This is great. You figured out how to pass custom callback to the `sort` method. It is imperative to note, however, that `sort` method alters the original array. As a result in this case the original data is no longer sorted by `id` as it used to be. 

A better strategy to have approached this challenge is to collect all car models in an array first and then sort this new array which contains only the strings of car models. The default behavior of sort is to sort alphabetically (make sure the array contains only text). So if we had collected all car models in the `carModels` array first, then in order to sort it all we would have had to do is to just call `carModels.sort()` without a callback. Putting the above-mentioned things into consideration you might end up with the following code.

```js
let carModels .= [];

for(let i = 0; i < inventory.length - 1; i++) {
  carModels.push(inventory[i].car_model);
}

console.log(carModels.sort());
```

ashwins93 (commented):
Great job. I love that you reset your progress and went through the challenge again. It is good to get some reps. I hope you enjoyed doing that and you picked up something new.

PR #449, opened 2018-10-15
Thomas Ketterman

Thomas-Ketterman:


Synch3:
Hi Thomas!

Nice work today, in our zoom meeting we discussed a couple different array methods and went over dot notation and template literals for objects.

PR #453, opened 2018-10-15
Justin Arata: Stretch Complete

jarata:


assignments/arrays.js
```diff
@@ -75,22 +81,34 @@ console.log();
...
 // ==== Challenge 4 ====
 // The accounting team needs all the years from every car on the lot. Create a new array from the dealer data containing only the car years and log the result in the console.
 let carYears = [];
-console.log();
+for (var i = 0; i < inventory.length; i++) {
+   carYears.push(inventory[i].car_year);
+}
+carYears.sort();
+
+console.log(carYears);
```
szincone, line 101: Could through the `sort()` into line 100 to save a line `console.log(carYears) => console.log(carYears.sort())`

assignments/arrays.js
```diff
@@ -75,22 +81,34 @@ console.log();
...
+}
+carModels.sort();
+
+console.log(carModels);
 
 // ==== Challenge 4 ====
 // The accounting team needs all the years from every car on the lot. Create a new array from the dealer data containing only the car years and log the result in the console.
 let carYears = [];
-console.log();
+for (var i = 0; i < inventory.length; i++) {
```
szincone: - please make sure you're not using `var` anymore. Use `let` or `const`. Since we're changing `i` everytime (adding to it) we would use `let`.

assignments/arrays.js
```diff
@@ -75,22 +81,34 @@ console.log();
...
+carYears.sort();
+
+console.log(carYears);
 
 // ==== Challenge 5 ====
 // The car lot manager needs to find out how many cars are older than the year 2000. Using the carYears array you just created, find out how many cars were made before the year 2000 by populating the array oldCars and logging it's length.
 let oldCars =[];
-console.log(); 
+for (var i = 0; i < inventory.length; i++) {
+   oldCars.push(inventory[i].car_year) >== 2000;
```
szincone: The ` >== 2000` isn't being read at all in this case. You'd want to do an if statement inside your forLoop and do something like `if (inventory[i].car_year >= 200){}` . You also only put one equal sign ` >== 2000` would be ` >= 2000`.

assignments/objects.js
```diff
@@ -18,31 +18,62 @@ const example = {
...
+console.log(interns[3]["name"]);
 // Antonietta's Gender
-
-// ==== Challenge 3: Object Methods ==== 
+console.log(interns[4]["gender"]);
+// ==== Challenge 3: Object Methods ====
 // Give Kennan the ability to say "Hello, my name is Kennan!" Use the console.log provided as a hint.
 // console.log(kennan.speak());
-
+console.log(kennan.speak());
```
szincone, line 70: - Where is the kennan object that you are trying to access? How did you access the previous interns? Why are you changing it for kennan?
- Does the kennan object have a speak method (function) inside of it?

assignments/arrays.js
```diff
@@ -65,6 +65,12 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
 console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
 
+for (let i = 0; i < inventory.length; i++) {
+    if (inventory[i]["id"] === 33) {
+        console.log(`Car 33 is a ${inventory[i]["car_year"]} ${inventory[i]["car_make"]} ${inventory[i]["car_model"]}`)
```
szincone: - Good use of template literals i all of your console.logs.

szincone (commented):
- Overall, really good job. It will get more familiar the more reps you get. Please check out my kennan questions and let me know if it doesn't make sense.

PR #454, opened 2018-10-15
Daren Larson

darenlarson:


assignments/arrays.js
```diff
@@ -63,34 +63,58 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 
+console.log('Car 33 is a ' + inventory[32].car_year + ' ' + inventory[32].car_make + ' ' + inventory[32].car_model);
 
 
 // ==== Challenge 2 ====
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
-let lastCar = 0;
-console.log();
+
+let lastCar = inventory.length - 1;
```
ashwins93, line 74: Good use of the `lastCar` variable, which caused a lot of confusion.

assignments/arrays.js
```diff
@@ -63,34 +63,58 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
...
 // ==== Challenge 3 ====
 // The marketing team wants the car models listed alphabetically on the website. Sort all the car model names into alphabetical order and log the results in the console
+
 let carModels = [];
-console.log();
+console.log(carModels);
+for(let i = 0; i < inventory.length; i++) {
+    carModels.push(inventory[i].car_model);
+}
+let sortedCarModels = carModels.sort();
```
ashwins93, line 86: `sort` function actually mutates the original array. What it means is that you don't need to store the return value of the sort function in another variable. As a matter of fact sort function returns the original array i.e. in this case `sortedCarModels === carModels` would return true.

assignments/objects.js
```diff
@@ -18,27 +18,73 @@ const example = {
 }
 
 // Write your intern objects here:
+const Mitzi = {
```
ashwins93, line 21: It is conventional to use `camelCase` for variable names. Why does it matter? Conventions allow for better communication between developers. If you see a camel-cased variable you know that the developer probably intended it to be the name of a function or a variable. `PascalCase` is used for naming special functions called *constructor functions*. You will learn more about this soon.

assignments/objects.js
```diff
@@ -18,27 +18,73 @@ const example = {
...
 // Give Kennan the ability to say "Hello, my name is Kennan!" Use the console.log provided as a hint.
-// console.log(kennan.speak());
+Kennan.speak = function() {
+  console.log("Hello, my name is " + Kennan.name +"!");
+};
+Kennan.speak();
 
 // Antonietta loves math, give her the ability to multiply two numbers together and return the product. Use the console.log provided as a hint.
 //console.log(antonietta.multiplyNums(3,4));
+Antonietta.multiplyNums = function(parameter1, parameter2) {
```
ashwins93, line 84: Using descriptive names instead of `parameter1` will add to better documentation.

ashwins93 (commented):
Good job. You hit the stretch. Keep it up.

PR #473, opened 2018-10-17
created intern objects

LeeTann:


abravebee:
# Objectives

# MVP
## Objects

  - [x] Challenge One
  - [x] Challenge Two
  - [ ] Challenge Three

## Arrays

  - [x] Challenge One
  - [ ] Challenge Two
  - [x] Challenge Three
  - [x] Challenge Four
  - [x] Challenge Five
  - [x] Challenge Six

# Stretch
## Tasks

# Great
- Excellent use of advanced Javascript syntax, and strong understanding of working with arrays and objects. Good job!!

# Requests for Improvement
- There are a few things that are very very nearly correct, with mistakes mainly in implementation, or overthinking. Unfortunately these do bring your rating down but they are not too complicated to fix: 
  - Objects Challenge Three: Your outputs are correct, but you were meant to create these functions outside the objects. When you want to add a property to an object outside of the object itself, you can simply declare it and set its value, and its value can be set to a function.
    - Example: `object.newProperty = function () { return or console.log something }`
  - Also, in the speak() function you have a console.log, and then you use a console.log when you call it (effectively `console.log((console.log(speak()))`), and apparently when you console.log a console.log, it adds an `undefined` in the console beneath the proper output. So just something to keep in mind, you'll most often want a return statement inside the function and a console.log when you're testing it. Otherwise, if your function contains a console.log, you can just call the function.
  - Arrays Challenge Two: You've grabbed the correct make and model, but your for loop is printing it 50 times, once for every item in the array, due to the for loop. You actually only needed the console.log for this one!
  - Arrays Challenge Six: Your logic is sound, but we want to return more than just the make, since that's the info we already know for sure. For example, you might have returned the year, make, and model of each car.

# Rating: 1

PR #476, opened 2018-10-17
Christene Bennett

christenebennett:
@abravebee

### DIFF
--- a/assignments/arrays.js
+++ b/assignments/arrays.js
@@ -7,7 +7,7 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 {"id":3,"car_make":"Land Rover","car_model":"Defender Ice Edition","car_year":2010},
 {"id":4,"car_make":"Honda","car_model":"Accord","car_year":1983},
 {"id":5,"car_make":"Mitsubishi","car_model":"Galant","car_year":1990},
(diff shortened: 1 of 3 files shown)


abravebee:
# Objectives

# MVP
## Objects

  - [x] Challenge One
  - [x] Challenge Two
  - [x] Challenge Three

## Arrays

  - [x] Challenge One
  - [x] Challenge Two
  - [x] Challenge Three
  - [x] Challenge Four
  - [x] Challenge Five
  - [x] Challenge Six

# Stretch
## Tasks

# Great
- You're demonstrating a solid understanding of objects and arrays here! Happy to see you did stretch as well! Well done!!

# Requests for Improvement
- I believe the intention of Objects Challenge 3 was to add the methods outside of the object. When you want to add a property to an object outside of the object itself, you can simply declare it and set its value, and its value can be set to a function.
  - Example: `object.newProperty = function () { return or console.log something }`
- Furthermore, lines 83 and 86 currently log `undefined`, as you are attempting to console.log a console.log. Since your test is a console.log, then you'll want a return in the function itself.
- Similar issue to above is occuring with your speak functions in the Objects stretch challenge, and the first Stretch Function Conversion challenge.
- Nothing wrong with your code or logic, but on Arrays Challenge 2, how could you use the `lastCar` variable provided to make the function itself a little more readable? Hint: you don't have to leave it set to 0.

# Rating: 3

PR #480, opened 2018-10-17
final

alexcartaz:


assignments/objects.js
```diff
@@ -19,26 +19,68 @@ const example = {
...
+}
+const antonietta = {
+	"id": 5,
+	"name": "Antonietta",
+  	"email": "adaine5@samsung.com",
+  	"gender": "F",
+  	"multiplyNums": function(a,b){
+  		return(a*b);
+  	}
+}
```
johnoro, line 57: Keep in mind that your object keys don't have to be strings (unless they require it due to spacing or something of that sort):
```javascript
const antonietta = {
	id: 5,
	name: "Antonietta",
  	email: "adaine5@samsung.com",
  	gender: "F",
  	multiplyNums: (a, b) => a * b
}
```

assignments/objects.js
```diff
@@ -49,7 +91,32 @@ const example = {
...
+			"speak": function(){
+		  		return("Hello, my name is " + this.name);
+		  	}
+		},
+		"speak": function(){
+			return("Hello, my name is " + this.name);
+		}
+	},
+	"speak": function(){
+  		return("Hello, my name is " + this.name);
```
johnoro, line 112: You could use a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) here:
```javascript
```

assignments/objects.js
```diff
@@ -19,26 +19,68 @@ const example = {
...
+  	"email": "mmelloy0@psu.edu",
+  	"gender": "F"
+}
+const kennan = {
+	"id": 2,
+	"name": "Kennan",
+  	"email": "kdiben1@tinypic.com",
+  	"gender": "M",
+  	"speak": function(){
+  		return("Hello, my name is Kennan!");
```
johnoro, line 34: You don't have to use the parentheses for the [`return`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/return) keyword:
```javascript
return "Hello, my name is Kennan!";
```

johnoro (commented):
You're missing `arrays.js`. Good work otherwise!

PR #495, opened 2018-10-18
Taylor Belk

TaylorBelk1:
@abravebee

### DIFF
--- a/assignments/arrays.js
+++ b/assignments/arrays.js
@@ -63,34 +63,63 @@ let inventory = [{"id":1,"car_make":"Lincoln","car_model":"Navigator","car_year"
 
 // ==== Challenge 1 ====
 // The dealer can't recall the information for a car with an id of 33 on his lot. Help the dealer find out which car has an id of 33 by logging the car's year, make, and model in the console log provided to you below:
-console.log(`Car 33 is a *car year goes here* *car make goes here* *car model goes here*` );
+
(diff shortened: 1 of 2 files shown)


abravebee:
# Objectives

# MVP
## Objects

  - [x] Challenge One
  - [x] Challenge Two
  - [x] Challenge Three

## Arrays

  - [x] Challenge One
  - [x] Challenge Two
  - [x] Challenge Three
  - [x] Challenge Four
  - [ ] Challenge Five
  - [x] Challenge Six

# Stretch
## Tasks

# Great
- Nearly perfect completion, just a couple small errors. Overall, you seem to be grasping the basics of working with arrays and objects. Well done!!

# Requests for Improvement
- On Objects Challenge 3, your logic is sound for adding the methods, but you don't have a way of checking this in the console. You can replace the console.logs inside each function with a return statement, and then console.log each function below (like the examples in the instruction). The code works fine when the functions are returned instead of logged, so that's the only issue there.
- In Arrays Challenge 5, you're unable to use `let carYears` on line 99 since it has already been declared and is being used in your function on line 91. If you remove 'let', it works just fine. You can also use a different variable name, or skip it altogether and simply remove line 103 and directly reference the car_year property on each inventory item in your if statement.
- For Arrays Challenge 6, we do want to use JSON.stringify() in our console.log per the instructions: `console.log(JSON.stringify(bmwAndAudi))`

# Rating:  2

PR #500, opened 2018-10-18
Zach Irvin JS - I

zacheryirvin:
@keirankozlowsk

### DIFF
--- a/assignments/arrays.js
+++ b/assignments/arrays.js
@@ -90,7 +90,7 @@ console.log(idToMakeModel(33));
 // The dealer needs the information on the last car in their inventory.  What is the make and model of the last car in the inventory?  Log the make and model into the console.
 function findLastCar() {
   const iPosistion = inventory.length - 1;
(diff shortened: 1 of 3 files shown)


keirankozlowski:
### Objective
The point of these assignments is to take your knowledge of JavaScript and start putting into practice the principles learned throughout JavaScript I.

### MVP
**Objectives**

- [x] Complete each challenge presented for Objects before moving on to Arrays.
- [x] Utilize the the array inventory to complete your challenges. Complete each challenge presented before moving on to stretch.

### Stretch
**Goals**

### Great
- All you work is solid! Great job!
- Stretch is mostly good.

### Requests for Improvement
- You've logged the child's name instead of age for the stretch question on the objects page.
- Instead of using charAt, you can use array methods on strings in JS. For example, you can use indexes.
- In stretch: `let anotherFunction = () => param;` swap "param" and the "()".

### Commits
- Great job!

### Rating: 3

PR #507, opened 2018-10-18
Javascript I

acarmstrong98:


abravebee:
# Objectives

# MVP
## Objects

  - [x] Challenge One
  - [x] Challenge Two
  - [ ] Challenge Three

## Arrays

  - [ ] Challenge One
  - [ ] Challenge Two
  - [x] Challenge Three
  - [x] Challenge Four
  - [ ] Challenge Five
  - [x] Challenge Six

# Stretch
## Tasks

# Great
- Most of your completed challenges look great, with just a few errors here and there.  Overall, you seem to be on the right track toward a stronger grasp of objects and arrays!

# Requests for Improvement
- It looks like you missed Objects Challenge 3, where you're meant to add methods to the Kennan and Antoinetta objects.
- In the Objects Stretch Challenge, you've forgotten to build the nested object that your return statements are attempting to reference. Also, we want `console.log`s to test our objects rather than return, unless you're invoking a function that includes a `console.log`.
- Double check what your console is logging for your Arrays Challenges. There are several that are not giving the correct outputs.  For example, your console.log on line 66 (Challenge 1) returns `car 33 is a [object Object]`. Think about the syntax you used in Challenges further down the list to grab specific info about each car. How can you grab just the year, make, and model of that object?
- Also remember to read the instructions carefully. For example, for Arrays Challenge 2, you've grabbed the right object, but the instructions only want you to log the make and model.

# Rating: 1

PR #512, opened 2018-10-20
Javascript I 

Pwolfe39:


abravebee:
# Objectives

# MVP
## Objects

  - [x] Challenge One
  - [x] Challenge Two
  - [x] Challenge Three

## Arrays

  - [ ] Challenge One
  - [ ] Challenge Two
  - [ ] Challenge Three
  - [ ] Challenge Four
  - [ ] Challenge Five
  - [ ] Challenge Six

# Stretch
## Tasks

# Great
- You’ve demonstrated a great understanding of working with objects! Arrays are unfinished so I don’t know exactly where you’re at with those—please reach out if you need help!

# Requests for Improvement
- It’s perfectly fine to have periods inside strings like in the example object. So for example, Keven’s email should log as `kmummery2@wikimedia.org`. You also have a few typos in your object values beside this, so just remember to be extra careful about these details.
- Looks like you forgot to log Antoinetta’s gender on line 66.
- Your logic for the object methods themselves is fine, but the intention here was to add the methods outside of the object. When you want to add a property to an object outside of the object itself, you can simply declare it and set its value, and its value can be set to a function.
  - Example: `object.newProperty = function () { return something }`
- No need to use sort in Array Challenge 1. What happened to the provided console.log there? You just need to grab the year, make, and model properties of inventory item number 33. In our coding challenges, we grab array items like so: `array[1]` where 1 is the item number. In the objects challenge, you grabbed properties with dot notation like so: `object.property`. How do you combine these two ideas to grab a property of a specific item in an array?

# Rating:  1

PR #524, opened 2018-11-26
completed challenge 1

siratl:
@fivegoblins "Project JS 1"

### DIFF
--- a/assignments/arrays.js
+++ b/assignments/arrays.js
@@ -2,7 +2,7 @@
 
 // The car dealer has all of their inventory housed in the array seen below.  Scroll down past the data to find out how you can help the car dealer.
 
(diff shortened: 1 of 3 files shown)


fivegoblins:
Great job, Elisha!

**Project Objectives**:

- **Objects Challenges**: Your objects are written correctly, your functions work and your console.log()s print the correct data to the console. Everything looks good here.

- **Arrays Challenges**: Every challenge except 2 was done correctly. The only issue with challenge 2 is that you won't always know how many items are in an array, so if you need to get the last item, you should assign your last item variable (in this case, lastCar) to : inventory[inventory.length - 1]. This way it will always grab the very last item no matter how many indexes or ids there are. Otherwise everything looked great!

**Stretch Goals**: You completed the stretch problem in the objects challenge and also did the arrow syntax exercise. 

**Rating: 3**

PR #528, opened 2018-11-26
JavaScript-I

kelliraehanson:


fivegoblins:
**Project Objectives**:

- **Objects Challenges**: You completed all of the challenges except the last one which was to give the Antoinetta object a function that multiplies two numbers.

-  **Arrays Challenges**: 

1. For Challenge 1 you did not console.log() the car_make or car_model properties for Car 33.
2. For Challenge 2 you should have assigned the variable lastCar to the last item in the inventory array, then logged both the make and model inside of one console.log().
3. For Challenge 3 you should have assigned the variable carModels to an empty array, then written a for loop that iterates over the inventory array and, at each index, pushes the car_model property into the empty carModels array, and then used .sort() to alphabetize.
4. Challenge 4 should have looked almost identical to Challenge 3.
5. For Challenge 5 you should have assigned the variable oldCars to an empty array, then written a for loop that iterates over the carYears array you created in Challenge 4. Your for loop should have contained an if statement that pushes indexes of carYears into your oldCars array IF they are less than 2000.
6. For Challenge 6 you had the correct answer but you commented it out!

**Rating: 1**

PR #564, opened 2018-11-26
objects and arrays

ghost:
@raylomeli


assignments/objects.js
```diff
@@ -33,15 +80,46 @@ const example = {
 
 // Antonietta's Gender
```
raylom, line 79: Consider reviewing the solution code for logging object data

PR #588, opened 2018-11-27
whatever

drkfogle:


VivaCode (commented):
#MVP

- [x] Objects
- [x] Arrays

#STRETCH

- [ ] Arrow Function Syntax
- [ ] Write some practice callbacks
- [ ] Refactor with array methods like .map(), .reduce(), .filter().

#COMMITS
Try to commit more frequently with professional and detailed messages.

#RATING: 2

PR #629, opened 2019-01-30
First commit object.js

Thacker-E:


Velsu (commented):
There is no solution to the exercise so i am unable to provide a proper code review.

